PR #2, opened 2022-11-09
mrc-3780 proof of concept

hillalex:
Implements the `GET /` endpoint, which just returns the outpack schema version.

Still kind of a lot of files but many of these are imported schemas.


src/main.rs
```diff
@@ -0,0 +1,51 @@
...
+        Ok(m) => { m }
+        Err(f) => {
+            print_usage(&program, opts);
+            panic!("{}", f.to_string())
+        }
+    };
+    Some(matches.opt_str("r").unwrap())
+}
+
+#[rocket::main]
```
richfitz, line 23: why does rocket get involved here?

src/main.rs
```diff
@@ -0,0 +1,51 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn can_parse_args() {
+        let root = parse_args(&[String::from("program"), String::from("--root"), String::from("test")]).unwrap();
+        assert_eq!(root, "test");
+
+        let root = parse_args(&[String::from("program"), String::from("-r"), String::from("test")]).unwrap();
```
richfitz, line 42: oh this `String::from` does not make me excited to do AoC in rust at all

tests/test_api.rs
```diff
@@ -0,0 +1,43 @@
...
+use std::path::Path;
+use rocket::local::blocking::Client;
+use rocket::http::{ContentType, Status};
+use jsonschema::{Draft, JSONSchema};
+
+#[test]
+fn can_get_index() {
+    let rocket = outpack_server::api(String::from("tests/example"));
+    let client = Client::tracked(rocket).expect("valid rocket instance");
+    let response = client.get("/").dispatch();
```
richfitz, line 11: this seems nice for testing the api

src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+
+#[derive(Responder)]
+#[response(status = 200, content_type = "json")]
+struct SuccessResponder<T> {
+    inner: Json<Response<T>>,
+    header: ContentType,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+struct ApiError {
```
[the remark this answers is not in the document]
richfitz: https://github.com/reside-ic/porcelain/blob/master/inst/schema/response-success.schema.json and https://github.com/reside-ic/porcelain/blob/master/inst/schema/response-failure.schema.json have the schema for this

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,12 @@
+on: push
+name: Clippy check
+jobs:
+  clippy_check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - run: rustup component add clippy
```
richfitz, line 8: does this action not need the toolchain action from above? Do they end up with the same version?

richfitz (changes requested):
Thanks for breaking this up, there's already heaps in here and it's pretty dense. Most of these comments are me wondering why things are done the way they are in rust I think not requests to change things. The only thing to change I think is the error response type

src/config.rs
```diff
@@ -0,0 +1,20 @@
...
+}
+
+pub fn read_config(root_path: &str) -> Result<Config, Error> {
+    let path = Path::new(root_path)
+        .join(".outpack")
+        .join("config.json");
+
+    let config_file = File::open(path)?;
+    let config: Config = serde_json::from_reader(config_file)?;
+    Ok(config)
```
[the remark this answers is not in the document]
hillalex: Yes. Should come right through to the API response. Let me add a test for this.

src/main.rs
hillalex: Yeah this multiple string types thing is horrible

src/lib.rs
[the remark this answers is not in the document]
hillalex: Doh, that was my intention, yes, the `ErrorResponder` was supposed to be using the `Response<T>` type

src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+
+#[macro_use]
+extern crate rocket;
+
+mod config;
+
+#[allow(clippy::result_large_err)]
+#[get("/")]
+fn index(root: &State<String>) -> Result<SuccessResponder<config::Config>, ErrorResponder> {
+    Ok(SuccessResponder::from(config::read_config(root)?))
```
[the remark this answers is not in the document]
hillalex: So the ErrorResponder is working here as expected I think, it's just that as you pointed out below I'm returning errors with the wrong schema.

src/main.rs
hillalex: https://api.rocket.rs/master/rocket/attr.main.html

Tbh I was just following a tutorial example, but now I see that it should be using the #launch macro instead: https://api.rocket.rs/master/rocket/attr.launch.html

src/lib.rs
```diff
@@ -0,0 +1,68 @@
+use std::io;
+use rocket::serde::{Deserialize, Serialize};
+use rocket::serde::json::{Json};
+use rocket::http::{ContentType};
+use rocket::{Build, Rocket};
+use rocket::State;
+
+#[macro_use]
```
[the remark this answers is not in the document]
hillalex: This is old syntax actually, I copied it from a tutorial and it imports all the rocket macros, but now they can be imported with `use` syntax so should probably remove this and import them that way.

tests/test_api.rs
hillalex: yeah it's really nice

src/main.rs
```diff
@@ -0,0 +1,51 @@
+use getopts::Options;
```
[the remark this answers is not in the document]
hillalex: Seemed ok, or as ok as parsing docopt ever is.

src/main.rs
hillalex: No, ok, there was a reason for this which is that #launch won't work unless all code paths return a rocket instance.

hillalex:
I've left the error-handling pretty crude - always returning a 500 rather than anything more specific. I think I need to try and find more examples of `rocket` error handling and come up with a full pattern here, so maybe one for a separate PR.

src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+pub fn api(root: String) -> Rocket<Build> {
+    rocket::build()
+        .manage(root)
+        .mount("/", rocket::routes![index])
+}
+
+#[derive(rocket::Responder)]
+#[response(status = 500, content_type = "json")]
```
richfitz, line 22: possibly we need a 400 responder here too:

```
$ curl -sv -H "Accept: application/json" localhost:8000/elsewhere  | jq
*   Trying ::1:8000...
* TCP_NODELAY set
* connect to ::1 port 8000 failed: Connection refused
*   Trying 127.0.0.1:8000...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8000 (#0)
> GET /elsewhere HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/7.68.0
> Accept: application/json
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 404 Not Found
< content-type: application/json
< server: Rocket
< permissions-policy: interest-cohort=()
< x-content-type-options: nosniff
< x-frame-options: SAMEORIGIN
< content-length: 128
< date: Thu, 17 Nov 2022 09:48:22 GMT
< 
{ [128 bytes data]
* Connection #0 to host localhost left intact
{
  "error": {
    "code": 404,
    "reason": "Not Found",
    "description": "The requested resource could not be found."
  }
}
```

(without the header it returns some html, which is fine if a bit gross, but the format here is unexpected)

richfitz: ah yes, I see in the logs here:

```
GET /elsewhere application/json:
   >> No matching routes for GET /elsewhere application/json.
   >> No 404 catcher registered. Using Rocket default.
   >> Response succeeded.
```

richfitz (changes requested):

src/lib.rs
richfitz: I figured that meant that on errors that we throw we'd return more sensible codes, but this I think belongs here - overriding the default 404 handler so that it matches the schema

hillalex:
Ok so this commit adds generic error handling for 404 and 500 errors (and removes the more specific handler I had in place for the index route; I'm still not 100% sure about the best way to achieve fine-grained error handling so that will be a further iteration)  https://github.com/mrc-ide/outpack_server/pull/2/commits/9d53f1f30cd33e47fc9907530c8eb4b79a5505d0

richfitz (approved):

PR #3, opened 2022-11-17
mrc-3794 add /metadata/list

hillalex:
Adds `GET /metadata/list` endpoint

### DIFF
--- a/schema/list.json
+++ b/schema/list.json
@@ -1,0 +1,20 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "array",
+  "items": {
+    "type": "object",
+    "properties": {
+      "packet": {
+        "$ref": "id.json"
+      },
+      "time": {
+        "type": "number"
+      },
+      "hash": {
+        "$ref": "hash.json"
+      }
+    },
+    "required": ["packet", "time", "hash"],
+    "additionalProperties": false
+  }
+}
--- a/schema/root.json
+++ b/schema/root.json
@@ -8,5 +8,6 @@
       "pattern": "^[0-9]+\\.[0-9]+\\.[0-9]+$"
     }
   },
-  "required": ["schema_version"]
+  "required": ["schema_version"],
+  "additionalProperties": false
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -2,19 +2,57 @@ use serde::{Deserialize, Serialize};
 use std::fs::File;
 use std::result::Result;
 use std::io::{Error};
-use std::path::{Path};
+use std::path::{Path, PathBuf};
+use cached::{Cached, cached_result};
 
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Location {
+    pub name: String,
+    pub id: String,
+    pub priority: i64,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct Config {
-    schema_version: String,
+    pub schema_version: String,
+    pub location: Vec<Location>,
+}
+
+cached_result! {
+    CONFIG_CACHE: cached::UnboundCache<PathBuf, Config> = cached::UnboundCache::new();
+    fn read_config_cached(path: PathBuf) -> Result<Config, Error> = {
+        let config_file = File::open(path)?;
+        let config: Config = serde_json::from_reader(config_file)?;
+        Ok(config)
+    }
 }
 
 pub fn read_config(root_path: &str) -> Result<Config, Error> {
     let path = Path::new(root_path)
         .join(".outpack")
         .join("config.json");
+    read_config_cached(path)
+}
 
-    let config_file = File::open(path)?;
-    let config: Config = serde_json::from_reader(config_file)?;
-    Ok(config)
+// to be used when changing config settings
+pub fn clear_cache() {
+    CONFIG_CACHE.lock().unwrap().cache_clear();
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn can_clear_cache() {
+        let _cfg = read_config("tests/example");
+        let _cached_cfg = read_config("tests/example");
+        assert_eq!(CONFIG_CACHE.lock().unwrap().cache_hits().unwrap(), 1);
+
+        clear_cache();
+
+        let _fresh_cfg = read_config("tests/example");
+        // should still only be 1 hit, since the last read should have been from disk
+        assert_eq!(CONFIG_CACHE.lock().unwrap().cache_hits().unwrap(), 1);
+    }
 }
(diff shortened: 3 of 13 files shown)


src/location.rs
```diff
@@ -0,0 +1,101 @@
...
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct LocationEntry {
+    pub packet: String,
+    pub time: f64,
+    pub hash: String,
+}
+
+const ID_REG: &str = "^([0-9]{8}-[0-9]{6}-[[:xdigit:]]{8})$";
+cached_result! {
+    ENTRY_CACHE: cached::UnboundCache<PathBuf, LocationEntry> = cached::UnboundCache::new();
```
hillalex, line 21: Caching here is ok because these will never change. This is an indefinite, in-memory cache.

src/config.rs
```diff
@@ -4,16 +4,36 @@ use std::result::Result;
 use std::io::{Error};
 use std::path::{Path};
 
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Serialize, Deserialize, Debug, Clone)]
```
richfitz, line 7: Why have we picked up `Clone` here? I'm sure there's a good reason it's just not jumping out at me from the code

src/lib.rs
```diff
@@ -1,11 +1,13 @@
...
-use rocket::{Build, catch, catchers, Request, Rocket};
+use rocket::{Build, catch, catchers, Request, Rocket, routes};
 use rocket::State;
 
 mod config;
 mod responses;
+mod location;
 
 use responses::{FailResponse, OutpackError, OutpackSuccess};
+pub use self::location::LocationEntry;
```
richfitz, line 10: Why does this need to be `pub use` and not `use`?

src/location.rs
```diff
@@ -0,0 +1,101 @@
...
+    let location_config = config::read_config(root_path)?.location;
+
+    let mut locations_sorted = fs::read_dir(&path)?
+        .into_iter()
+        .filter_map(|r| r.ok())
+        .collect::<Vec<DirEntry>>();
+
+    locations_sorted.sort_by(|a, b| get_priority(&location_config, a).cmp(&get_priority(&location_config, b)));
+
+    let reg = Regex::new(ID_REG).unwrap();
```
richfitz, line 68: if this was `id_reg` throughout it would make the connection to `ID_REG` clearer

src/location.rs
```diff
@@ -0,0 +1,101 @@
...
+
+fn is_packet(name: &OsString, reg: &Regex) -> bool {
+    let o = name.to_str();
+    o.map_or(false, |s| reg.is_match(s))
+}
+
+fn get_priority(location_config: &Vec<Location>, entry: &DirEntry) -> i64 {
+    let id = entry.file_name();
+    location_config.into_iter()
+        .find(|l| OsString::from(&l.id) == id)
```
richfitz, line 37: We need OsString here because we're interacting with things passed back from the system calls about the file system?

src/lib.rs
```diff
@@ -26,15 +28,23 @@ fn not_found(_req: &Request) -> Json<FailResponse> {
...
 #[rocket::get("/")]
-fn index(root: &State<String>) -> OutpackResult<config::Config> {
+fn index(root: &State<String>) -> OutpackResult<config::Root> {
     config::read_config(root)
+        .map(|r| config::Root::new(r.schema_version))
+        .map_err(|e| OutpackError::new(e))
+        .map(|r| OutpackSuccess::from(r))
+}
+
+#[rocket::get("/metadata/list")]
```
richfitz, line 38: needs adding to the README with an example

src/location.rs
```diff
@@ -0,0 +1,101 @@
...
+        .collect::<Vec<DirEntry>>();
+
+    locations_sorted.sort_by(|a, b| get_priority(&location_config, a).cmp(&get_priority(&location_config, b)));
+
+    let reg = Regex::new(ID_REG).unwrap();
+
+    let packets = locations_sorted
+        .into_iter()
+        .map(|entry| read_location(entry.path(), &reg))
+        .into_iter()
```
richfitz, line 73: why do we need all these into_iter entries? can we not just map/collect directly? I presume not given you've done it this way, but what is the language trying to do here?

src/config.rs
```diff
@@ -4,16 +4,36 @@ use std::result::Result;
...
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct Config {
-    schema_version: String,
+    pub schema_version: String,
+    pub location: Vec<Location>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Root {
```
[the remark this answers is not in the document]
richfitz: What is a DTO sorry?

schema/root.json
```diff
@@ -8,5 +8,6 @@
       "pattern": "^[0-9]+\\.[0-9]+\\.[0-9]+$"
     }
   },
-  "required": ["schema_version"]
+  "required": ["schema_version"],
+  "additionalProperties": false
```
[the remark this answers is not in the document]
richfitz: actually this is totally different to the root schema there:

https://github.com/mrc-ide/outpack.server/blob/main/inst/schema/root.json

richfitz (changes requested):
Mostly comments really!

src/location.rs
hillalex: Yeah I had the same questions. Probably some sized cache makes sense, boots out the oldest entries: https://docs.rs/cached/latest/cached/stores/struct.SizedCache.html But wasn't sure what size to go for?

src/config.rs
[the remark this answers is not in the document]
hillalex: Data Transfer Object (MVC terminology!) - models that exist purely to return data to the client.

schema/list.json
```diff
@@ -0,0 +1,20 @@
+{
```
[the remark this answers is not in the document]
hillalex: The downloader script is atm just pulling `outpack` schemas; it could pull schemas from the R `outpack.server`, but it wasn't clear to me that that's the canonical place for the api schemas to live as opposed to this repo?

src/lib.rs
hillalex: Looks vestigial, probably was because I was referencing that type in integration tests, but am not any longer so can remove this whole statement :+1:

src/location.rs
hillalex: yeah, we're comparing against `entry.file_name()` which returns an `OsString`. could also cast the other way, should be totally innocent either way in this case

src/location.rs
hillalex: Huh, no, you're totally right, some of these are superfluous and I can't tell you why other than I fiddled around with this code a lot :see_no_evil:

schema/list.json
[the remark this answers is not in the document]
richfitz: absolutely. Perhaps we need an schemas repo to be the single source of truth for everything?

src/location.rs
richfitz: Let's leave as you have it for now and see how it goes

richfitz (approved):

PR #4, opened 2022-11-23
mrc-3828

hillalex:


src/responses.rs
```diff
@@ -1,11 +1,13 @@
...
-use rocket::http::{ContentType};
+use rocket::serde::json::{Json, json};
+use rocket::http::{ContentType, Status};
+use rocket::{Request, Response};
 use rocket::response::Responder;
 
 #[derive(Responder)]
 #[response(status = 200, content_type = "json")]
-pub struct SuccessResponder<T> {
+pub struct OutpackSuccess<T> {
```
hillalex, line 10: renamed for symmetry with `OutpackError`, which now also implements the `rocket::Responder` trait (which all endpoints need to return)

src/responses.rs
```diff
@@ -16,6 +18,26 @@ pub struct OutpackError {
     pub detail: String,
 }
 
+impl OutpackError {
+    pub fn new(e: io::Error) -> OutpackError {
```
hillalex, line 22: So we'd have to extend this whenever a known error of a different type could have occurred

src/lib.rs
```diff
@@ -1,12 +1,13 @@
...
 use rocket::{Build, catch, catchers, Request, Rocket};
 use rocket::State;
 
 mod config;
 mod responses;
 
-use responses::{FailResponse, OutpackError, SuccessResponder};
+use responses::{FailResponse, OutpackError, OutpackSuccess};
+
+type OutpackResult<T> = Result<OutpackSuccess<T>, OutpackError>;
```
hillalex, line 10: so all endpoints should return something of this type

src/responses.rs
```diff
@@ -16,6 +18,26 @@ pub struct OutpackError {
...
+        OutpackError {
+            error: e.kind().to_string(),
+            detail: e.to_string()
+        }
+    }
+}
+
+impl<'r> Responder<'r, 'static>  for OutpackError {
```
richfitz, line 30: ok, what is going on with all these apostrophes here, this looks very odd

richfitz (approved):
OK, this looks good but very dense - I guess we'll get used to that!

Any pointers to what is going on with the apostrophes would be interesting to read

src/responses.rs
hillalex: these are lifetimes, but don't ask me why this needs annotating or whether `'static` is the best choice here :see_no_evil: https://doc.rust-lang.org/book/ch10-03-lifetime-syntax.html

PR #6, opened 2022-12-14
mrc-3883 add GET file endpoint

hillalex:
This was branched off the previous PR so shouoldn't be merged yet but can be reviewed.


src/responses.rs
```diff
@@ -19,21 +19,20 @@ pub struct OutpackError {
     pub detail: String,
 
     #[serde(skip_serializing, skip_deserializing)]
-    pub kind: Option<ErrorKind>
+    pub kind: Option<ErrorKind>,
 }
 
 impl OutpackError {
-    pub fn new(e: io::Error) -> OutpackError {
+    pub fn from(e: io::Error) -> OutpackError {
```
hillalex, line 26: This is unrelated but `from` seems more correct and we might reserve `new` for directly creating an error object by passing explicit `kind`, `error` and `detail` constructor arguments.

hillalex:
this is ready for review/merge

src/store.rs
```diff
@@ -0,0 +1,42 @@
+use std::path::{Path, PathBuf};
+use regex::Regex;
+
+const HASH_REG: &str = "^([[:alnum:]]+):([[:xdigit:]]+)$";
+
+pub fn hash_parse(hash: &str) -> (&str, &str) {
+    let hash_reg = Regex::new(HASH_REG).unwrap();
+    let caps = hash_reg.captures(hash).unwrap();
+    (caps.get(1).map_or("", |m| m.as_str()),
```
richfitz, line 9: no test below of the `""` behaviour here - worth testing? worth returning a Result type?

src/store.rs
```diff
@@ -0,0 +1,42 @@
+use std::path::{Path, PathBuf};
+use regex::Regex;
+
+const HASH_REG: &str = "^([[:alnum:]]+):([[:xdigit:]]+)$";
+
+pub fn hash_parse(hash: &str) -> (&str, &str) {
```
richfitz, line 6: Is it worth returning a struct rather than tuple, to improve readability? We might not do so much with these, that it matters?

richfitz (commented):
A few small comments or suggestions, none need doing though!

src/outpack_file.rs
```diff
@@ -0,0 +1,44 @@
...
+    size: u64,
+}
+
+impl OutpackFile {
+    pub async fn open<P: AsRef<Path>>(hash: String, path: P) -> io::Result<OutpackFile> {
+        let file = File::open(path.as_ref()).await.map_err(|e| {
+            return match e.kind() {
+                ErrorKind::NotFound => io::Error::new(ErrorKind::NotFound,
+                                                      format!("hash '{}' not found", hash)),
+                _ => e
```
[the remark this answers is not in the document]
hillalex: Yeah, I'm not sure what else in practice with this call, but there is a long list of possible `ErrorKind` https://doc.rust-lang.org/std/io/enum.ErrorKind.html

src/store.rs
hillalex: Yeah maybe better to return `Result` actually, since otherwise you'll just get an error at the next step with a message that doesn't make loads of sense. Will change over and test :+1:

richfitz (approved):

PR #7, opened 2022-12-19
add docker file etc

hillalex:


Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM rust:latest as builder
+WORKDIR /usr/src/outpack_server
+COPY . .
+RUN cargo install --path .
+
+FROM debian:buster-slim
+COPY --from=builder /usr/local/cargo/bin/outpack_server /usr/local/bin/outpack_server
```
richfitz, line 7:
```suggestion
COPY --from=builder /usr/local/cargo/bin/outpack_server /usr/local/bin/outpack_server
EXPOSE 8000
```

richfitz (changes requested):
One suggestion otherwise good to go

richfitz (approved):
nice, all working for me

These docker images are nice and small, too (or comparatively at least - 80MB which is what you'd expect from size of the debian:buster-slim image + the 11MB exe)

PR #9, opened 2023-02-28
Reorganise source for eventual cli app

richfitz:
This PR reorganises the source - this builds on the pattern that was already most there.

* we now build two binaries - `outpack_server` and `outpack`; the former is the http api and the latter is a smaller cli app that we will expand
* moved the api code into its own file and reorganised the lib.rs file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
 .idea
 montagu-reports
+tests/example/.outpack/index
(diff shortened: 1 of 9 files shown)


r-ash (approved):
Looks good!

hillalex (approved):
nice :+1:

PR #10, opened 2023-04-12
mrc-4154: Add skeleton querying CLI tool

r-ash:
This PR will
* Add a new CLI `outpack_query` which takes root path and a query and searches the index with the query returning the result in plain text

Note that at the moment only query `latest` and lookup for an ID works, other queries will error. 

This PR adds an `Index` struct which is a collection of `Packet`s. At the moment the `Packet` only has an `id` but I'm expecting to expand this in future PRs with `name`, `parameters`, `location` and dependency information. 

The querying works by
1. Parsing the query string using [pest](https://pest.rs/book/intro.html) into a `QueryNode` object (the `QueryNode` is our abstract syntax tree)
2. Evaluating the query from the `QueryNode`

My plan for expanding this will be `QueryNode` will get new variants some of which might have a `QueryNode` themselves, e.g. for `latest` as a function will be `Latest(QueryNode)` or a grouping will be `DyadicGrouping(GroupType, QueryNode, QueryNode)` or `MonadicGrouping(GroupType, QueryNode)` where ` GroupType` will be an enum with `Intersection`, `Union`, `Negation`

Next steps
* Support latest as a function which takes 0 or 1 arg
* Support lookups by name (and `==` test)
* Support returning multiple results (at the moment it will only return 1 ID)
* Support searching for parameters
* Support groupings `(` , `!`, `&&` and `||`
* Support remaining tests `!=`, `<`, `<=`, `>`, `>=`
* Support `single`
* Support searching location with `at_location`
* Support subqueries and dependencies
* Update CLI to be more git-like e.g. `outpack --root <path> query <query>` or similar
* Improve errors


src/index.rs
```diff
@@ -0,0 +1,33 @@
+use std::io;
```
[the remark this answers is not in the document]
r-ash: Yep, will definitely grow a bunch

src/query.rs
```diff
@@ -0,0 +1,98 @@
...
+    eval_query(index, parsed)
+}
+
+#[derive(Parser)]
+#[grammar = "query.pest"]
+pub struct QueryParser;
+
+enum QueryNode<'a> {
+    Latest,
+    Lookup(&'a str)
```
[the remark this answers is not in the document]
r-ash: yes I think so, will try and keep it similar where it makes sense and where it doesn't feel light fighting against Rust. Will especially try and keep the terminology the same as from the R impl

tests/test_query.rs
```diff
@@ -0,0 +1,58 @@
...
+fn eval_error_can_be_displayed() {
+    let err = QueryError::EvalError("my error msg".to_string());
+    let text = format!("{}", err);
+    assert_eq!(text, "Failed to evaluate query\nmy error msg");
+}
+
+#[test]
+fn can_get_packet_by_id() {
+    let root_path = "tests/example";
+    let ids = outpack::query::run_query(root_path, "\"20170818-164847-7574883b\"".to_string())
```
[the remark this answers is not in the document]
r-ash: Yeah it feels a bit strange, not sure how to work around it though. Currently using the fact that the text is quoted to indicate that this is a string. So outer `'` passes it to the query, inner `"` used to show that the following characters are a string. I think makes more sense when you think about it in the context of a longer query which will come later
```
--q 'id == "20170818-164847-7574883b"'
```
or 
```
--q 'latest("20170818-164847-7574883b")'
```

In R it needs to be quoted too
```
> outpack_query(quote("20230419-144558-59cc3c21"), root = root)
[1] "20230419-144558-59cc3c21"
> outpack_query(quote(20230419-144558-59cc3c21), root = root)
Error: unexpected symbol in "outpack_query(quote(20230419-144558-59cc3c21"
```

but we don't have issues with it as the query itself doesn't get passed as a string.

I feel like that is fairly normal for CLI tools though? Using R on CLI if you want to just send a string its `R -e '"thing"'`. I'm not sure what else we would do here, do you have any good ideas?

src/query.rs
```diff
@@ -0,0 +1,98 @@
...
+    }
+}
+
+fn eval_latest(index: Index) -> Result<String, QueryError> {
+    Ok(index.packets.last().unwrap().id.clone())
+}
+
+fn eval_lookup(index: Index, value: &str) -> Result<String, QueryError> {
+    let ids: Vec<String> = index.packets.into_iter().map(|packet| packet.id).collect();
+    let exists = ids.iter().any(|id| id == value);
```
hillalex, line 72: is there a reason to collect ids into a vector and then back into an iterator to check if any match? seems like this would be more efficient:
`let exists = index.packets.into_iter().any(|packet| packet.id == value);`

src/query.rs
```diff
@@ -0,0 +1,98 @@
...
+      }
+  }
+}
+
+fn parse_query(query: &str) -> Result<QueryNode, QueryError> {
+    match QueryParser::parse(Rule::query, query) {
+        Ok(mut pairs) => {
+            // Below never fails as query has been parsed and we know query rule can only have 1
+            // expr and its inner can only be length 1 also (either latest or a string)
+            let query = pairs.next().unwrap().into_inner().next().unwrap();
```
hillalex, line 45: I was a bit confused by the calls to `next` here. If there is only one each time, how about `peek` intead? (then also lose the `mut` on line 42)
```suggestion
            let query = pairs.peek().unwrap().into_inner().peek().unwrap();
```

Similarly on line 49

src/index.rs
```diff
@@ -0,0 +1,33 @@
...
+use crate::metadata::get_ids;
+
+pub struct Index {
+    pub packets: Vec<Packet>
+}
+
+pub struct Packet {
```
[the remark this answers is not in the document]
r-ash: I think we need it to be public so that we can loop over it in query.rs. Otherwise getting error like
```
error: type `Packet` is private
  --> src/query.rs:67:8
   |
67 |     Ok(index.packets.last().unwrap().id.clone())
   |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ private type
```

Though yeah I think you're right, I'm not sure having a `Packet` object is helping at moment. Probably some more efficient way to represent that in the index. Will have a think

src/index.rs
```diff
@@ -0,0 +1,33 @@
+use std::io;
+use crate::metadata::get_ids;
+
+pub struct Index {
+    pub packets: Vec<Packet>
```
hillalex, line 5: this is a struct because down the line we'll want other info aside from just ids?

r-ash: Yeah, so if we follow same structure as in R we'll also store here some data structure for the dependencies between packets. And a `Packet` itself will have info about name, parameters and location

src/query.rs
r-ash: Thanks! Fixed this up

src/query.rs
r-ash: Thanks, fixed this up

src/index.rs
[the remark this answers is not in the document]
r-ash: Ok having a look around I think this is probably a good way to start trying to encapsulate the filtering logic into one place. I think probably I'd like to leave this as is for now and if we see the querying is slow we can look at some other approaches.

richfitz (approved):

src/query.rs
```diff
@@ -0,0 +1,104 @@
...
+pub struct QueryParser;
+
+enum QueryNode<'a> {
+    Latest,
+    Lookup(&'a str),
+}
+
+#[derive(Debug, Clone)]
+pub enum QueryError {
+    ParseError(Box<pest::error::Error<Rule>>),
```
hillalex, line 31: maybe worth a comment about why a Box is needed here? I don't really get it :sweat_smile:

hillalex (approved):

src/query.rs
r-ash: Thanks, added that

PR #12, opened 2023-05-17
mrc-4249: Add name lookup and inner query to latest()

r-ash:
The PR is a little noisy because I ran rustfmt on the querying files and split the querying impl into separate files within a submodule. I can back that out for another PR if it is a bit too much.

At the moment we can run queries like
* `"123423"` - to search for that ID
* `latest` - to find the latest

This PR will add support to query CLI so that we can run queries like
* `id == "123123142123"` - to search for ID explicitly
* `name == "thing"` - to search for a name
* `latest()` - to get the latest
* `latest(name == "thing")` - to get the latest where `name == "thing"`

Made a few structural changes to support this including
* Moved the querying into a submodule. This has publicly exported
   * `run_query`
   * `QueryError` - not sure if we need this, it's exported publicly atm because I am using it in the integration tests but not sure what is standard is Rust. Libs probably do export the type of any potential errors?
* Splits up the different parts of query running in the submodule to now have a separate file for query parsing, query evaluation and formatting.
* Use the Packet type from metadata module
* Updates some of the pest query parsing language. Their website has a neat way to visualise the result of running the pest parsing. See https://pest.rs/#editor. For us here are some example queries and their resulting parsed structure
   * "12345"
      ```
      - query
        - string > inner: "12345"
        - EOI: ""
      ```
   * latest or latest()
      ```
      - query
        - noVariableFunc > latest: "latest"
        - EOI: ""
      ```
   * name == "example"
      ```
      - query
        - infixExpression
          - firstArg: "name"
          - infixFunction: "=="
          - secondArg > string > inner: "example"
        - EOI: ""
      ```
   * latest(name == "example")
      ```
      - query
        - singleVariableFunc
          - latest: "latest"
          - arg > infixExpression
            - firstArg: "name"
            - infixFunction: "=="
            - secondArg > string > inner: "example"
        - EOI: ""
      ```

Some things I'd like to tidy up from this, but perhaps down the line
* Where do we return String vs `&str`, I do a fair bit of cloning here of IDs, which should be fairly small amount of data but should I put some effort into trying to use the reference instead where I can
* Improving the error handling, and deciding where the split should be. Here the pest language is fairly prescriptive in some ways e.g. any function of the form `func_name()` the function name can only be `latest` and pest will reject anything else. But for infix operators pest will take any and we reject later in the code if it is not one we recognise. Similarly returning more information when there are no packets found.


src/query/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+    format_query_result(result)
+}
+
+#[derive(Debug, Clone)]
+// Results with QueryError are at least as large as the QueryError variant. The compiler
+// will need to reserve that much memory every time it is used. We want to keep this as
+// small as possible so Box the large error body to force it onto the heap.
+// See https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
+pub enum QueryError {
+    ParseError(Box<pest::error::Error<Rule>>),
```
richfitz, line 35: Add a constructor that will box the error to avoid having all the Box's throughout here (if that is possible)

src/query/query.pest
```diff
@@ -0,0 +1,24 @@
+query         =  { SOI ~ interfaceExpr ~ EOI }
+interfaceExpr = _{ expr | string }
+// Interface expr can a standard expression or literal ID
+expr = _{ singleVariableFunc | noVariableFunc | infixExpression }
+
+noVariableFunc     =  { funcNames ~ "()" | latest }
```
richfitz, line 6: Do some basic prep beforehand to process literal ids into `single(id == 'literal')` and `latest` into `latest()`, then simplify the grammar

src/query/query.pest
```diff
@@ -0,0 +1,24 @@
+query         =  { SOI ~ interfaceExpr ~ EOI }
```
richfitz, line 1: Add a link here to the pest playground

src/query/query_eval.rs
```diff
@@ -0,0 +1,76 @@
...
+        QueryNode::Latest(inner) => eval_latest(index, inner),
+        QueryNode::Lookup(field, value) => eval_lookup(index, field, value),
+    }
+}
+
+fn eval_latest(index: Index, inner: Option<Box<QueryNode>>) -> Result<Vec<Packet>, QueryError> {
+    match inner {
+        Some(inner_node) => Ok(vec![eval_query(index, *inner_node)?
+            .last()
+            .unwrap()
```
richfitz, line 17: There was a hint in the rust book about unwrap with a self-documenting reason on failure

src/query/query_parse.rs
```diff
@@ -0,0 +1,131 @@
...
+                        _ => unreachable!(),
+                    };
+                    let search_term = get_string_inner(second_arg.into_inner().peek().unwrap());
+                    Ok(QueryNode::Lookup(lookup_type, search_term))
+                }
+                _ => {
+                    let err = pest::error::Error::new_from_span(
+                        pest::error::ErrorVariant::CustomError {
+                            message: format!(
+                                "Encountered unknown infix operator: {}",
```
richfitz, line 52:
```
cargo run --bin outpacl_query -- -r tests/example/ -q 'name != 123'
```

does not actually throw this - possibly the parser is swallowing it. Code coverage would make this easier to test

richfitz (changes requested):
Also might be worth some generic error wrapper around pest's errors to smooth over any differences

src/query/query_eval.rs
```diff
@@ -0,0 +1,76 @@
...
+use crate::query::QueryError;
+
+pub fn eval_query(index: Index, query: QueryNode) -> Result<Vec<Packet>, QueryError> {
+    match query {
+        QueryNode::Latest(inner) => eval_latest(index, inner),
+        QueryNode::Lookup(field, value) => eval_lookup(index, field, value),
+    }
+}
+
+fn eval_latest(index: Index, inner: Option<Box<QueryNode>>) -> Result<Vec<Packet>, QueryError> {
```
hillalex, line 13: So this always returns a vector of length 1 but has to be a vector to match the return type of `eval_lookup`?

src/query/query_eval.rs
hillalex, line 17: unwrapping always fine here? alternative would be `ok_or(some error)`

src/query/query_parse.rs
```diff
@@ -0,0 +1,131 @@
...
+        }
+        Rule::noVariableFunc => {
+            // Pest asserts for us that the only possible no variable func is latest()
+            // we might want to move this validation into Rust code later to return
+            // better errors to the user
+            Ok(QueryNode::Latest(None))
+        }
+        Rule::infixExpression => {
+            let mut infix = query.into_inner();
+            let first_arg = infix.next().unwrap();
```
hillalex, line 35: lots of `unwrap` here, which I feel like is something we're supposed to avoid. how about `.map(|e| e.as_str())` instead and then use `Some("id")` etc on line 41 etc? and similarly for other options?

hillalex (changes requested):
I think there might be some more idiomatic ways to handle the options (rather than unwrapping them)?

src/query/mod.rs
r-ash: I think no straightforward enum variant ctor. I could write another function to build this from the `Rule` but then that is passing the unboxed variable into a function, but then I think (?) Rust will potentially copy the potentially large `pest::error::Error` when moving it into the function. After reading through this weeks Rust chapter it looks like this is the idiomatic way to box a variant on an enum.

src/bin/outpack_query.rs
```diff
@@ -28,10 +28,12 @@ fn main() {
...
     let result = outpack::query::run_query(&root, query);
     match result {
-        Ok(res) => println!("{}", res),
+        Ok(res) => {
+            println!("{}", res)
+        }
         Err(e) => {
             eprintln!("{}", e);
             process::exit(1);
-        },
```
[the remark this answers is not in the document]
r-ash: Ah so this is being removed in IntelliJ by rustfmt. What do you tend to use here? Just clippy?

src/query/mod.rs
```diff
@@ -0,0 +1,46 @@
+mod query_parse;
+mod query_types;
+mod query_eval;
+mod query_format;
+
+extern crate pest;
+
+use crate::index::get_packet_index;
+use std::fmt;
```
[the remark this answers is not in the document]
r-ash: Just ran rustfmt on this file, that is putting crate imports first

src/query/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+use crate::query::query_parse::{parse_query, Rule};
+use crate::query::query_eval::eval_query;
+use crate::query::query_format::format_query_result;
+
+pub fn run_query(root: &str, query: String) -> Result<String, QueryError> {
```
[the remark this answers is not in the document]
r-ash: Thanks - fixed this up

src/query/query_eval.rs
r-ash: Thanks, I wasn't handling this case! Have added a test now and reworked this code

src/query/query_eval.rs
```diff
@@ -0,0 +1,76 @@
...
+    match inner {
+        Some(inner_node) => Ok(vec![eval_query(index, *inner_node)?
+            .last()
+            .unwrap()
+            .clone()]),
+        None => Ok(vec![index.packets.last().unwrap().clone()]),
```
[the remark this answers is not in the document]
r-ash: Yep! Have updated this to remove the number of clones. The evaluation now takes a reference to index and returns a reference to the packet from that index. Should hopefully mean less copying. Have reworked this code to handle that

src/query/query_parse.rs
r-ash: Thanks, fixed this now. You should now get

```
cargo run --bin outpack_query -- -r tests/example/ -q "name != \"123\""
```

```
Failed to parse query
 --> 1:6
  |
1 | name != "123"
  |      ^^
  |
  = Encountered unknown infix operator: !=

Process finished with exit code 1

```

src/query/query.pest
r-ash: Thanks, have tidied this up

src/query/query_parse.rs
r-ash: In the pest docs (https://pest.rs/book/parser_api.html#using-pair-and-pairs-with-a-grammar) it says 

> you should rely on the meaning of your grammar for properties such as "contains n sub-rules", "is safe to parse to f32", and "never fails to match". Idiomatic pest code uses unwrap and unreachable!.

I can update the parsing to remove usages of unwrap but I think they are safe in the parsing because `pest` ensures that those things are available. I'm happy update this to not use `unwrap` if you feel strongly but I think it is not as bad in this context as it is in other code.

r-ash:
> Also might be worth some generic error wrapper around pest's errors to smooth over any differences

You mean by like pulling out the message from the pest error and just returning some simple `Error` type with a message? I was reluctant to do this because then we'd potentially lose info which might be valuable later. Also using the pest error types means we get their nice printing like in example here https://github.com/mrc-ide/outpack_server/pull/12#discussion_r1226638494

richfitz (approved):

hillalex (approved):

PR #13, opened 2023-06-14
mrc-4175 add missing ids endpoint

hillalex:
Adds the endpoint for determining missing packet ids. Intended as an equivalent to the following implementation in the file location:

```
root_list_unknown_packets <- function(ids, unpacked, root) {
  if (unpacked) {
    setdiff(ids, root$index()$unpacked$packet)
  } else {
    setdiff(ids, names(root$index()$metadata))
  }
}
```


r-ash (approved):
Nice, looks good

src/metadata.rs
```diff
@@ -112,6 +120,24 @@ pub fn get_ids(root_path: &str) -> io::Result<Vec<String>> {
...
+    let s = String::from(id.trim());
+    if is_packet_str(&s) {
+        Ok(s)
+    } else {
+        Err(io::Error::new(io::ErrorKind::InvalidInput,
+                           format!("Invalid packet id '{}'", id)))
+    }
+}
+
+pub fn get_missing_ids(root_path: &str, wanted: &str, unpacked: Option<bool>) -> io::Result<Vec<String>> {
```
richfitz, line 133: I think that for the metadata functions, this should be taking something like `&[&str]` or `&[String]` with the API part taking responsibility for the split

richfitz (commented):
One query and one design idea. Leaving this as comment so it can easily be meregd while I am away

PR #15, opened 2023-06-19
mrc-4335: Add parameters search into outpack rust querying

r-ash:
Note this will only work for string types at the moment, will add querying for numeric types in a separate PR.

But after this PR you can now run a query like
```
parameter:disease == "YF"
```

or 

```
latest(parameter:iso3 == "MWI")
```

### DIFF
--- a/src/query/query_eval.rs
+++ b/src/query/query_eval.rs
@@ -30,6 +30,19 @@ fn eval_latest<'a>(
     }
 }
 
+fn packet_has_param(packet: &Packet, key: &str, value: &str) -> bool {
+    match &packet.parameters {
+        Some(params) => {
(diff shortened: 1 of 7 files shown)


src/query/query_eval.rs
```diff
@@ -30,6 +30,19 @@ fn eval_latest<'a>(
     }
 }
 
+fn packet_has_param(packet: &Packet, key: &str, value: &str) -> bool {
+    match &packet.parameters {
+        Some(params) => {
+            if let Some(existing_value) = params.get(key) {
+                *existing_value == value
```
hillalex, line 37: Not sure this asterisk is needed?

r-ash: Thanks! It is not, I've removed that now

PR #16, opened 2023-06-21
change entrypoint

hillalex:


start-with-wait
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -e
+
+until [ -f /outpack/.outpack/config.json ]; do
+  sleep 3
+done
```
r-ash, line 6: Could this loop forever if something goes wrong? Maybe worth putting a max number of retries on. There's a script which does that in hintr if it is useful https://github.com/mrc-ide/hintr/blob/master/docker/test_connection#L45

r-ash (commented):
One small comment - it might be so low a risk not worth worrying about

PR #17, opened 2023-06-21
mrc-4309 endpoint to list missing files

hillalex:
Implements `POST /files/missing` (as with missing packets, a POST rather than a GET to avoid running into limits with query string length)

Equivalent of this function for the file remote, but since http remotes must always use a file store, only the logic under the first part of the if clause is needed.
```
root_list_unknown_files <- function(hashes, root) {
  if (root$config$core$use_file_store) {
    hashes[!root$files$exists(hashes)]
  } else {
    idx <- root$index()
    if (length(idx$unpacked$packet) == 0) {
      return(hashes)
    }
    ## This could be quite a slow operation, especially if we always
    ## validate each file (as we currently do)
    hashes[vlapply(hashes, function(h) is.null(find_file_by_hash(root, h)))]
  }
}
```


tests/test_api.rs
```diff
@@ -321,6 +376,7 @@ fn validate_error(instance: &Value, message: Option<&str>) {
             .expect("Error detail")
             .to_string();
 
+        print!("{}", err);
```
r-ash, line 379: Is this left over from some debugging?

r-ash (approved):
This looks great, one small comment about a stray print line

tests/test_api.rs
hillalex: Thanks, changed to a custom panic message instead.

PR #19, opened 2023-06-26
mrc-4176 post file method

hillalex:
Adds `POST file/hash` which uploads a file to the server (after validating the hash).


README.md
```diff
@@ -250,6 +250,23 @@ Given a list of file hashes, returns those that are missing in the current root.
...
+### Body
+The file contents should be written directly to the request body.
+
+### Response
+Returns the location of the file on the server.
+```
+{
+  "status": "success",
+  "errors": null,
+  "data": "/outpack/.outpack/files/sha256/b1/89579a9326f585d308304bd9e03326be5d395ac71b31df359ab8bac408d248"
```
richfitz, line 266: why do we return this? when would a consumer need to know where the path was? I would have thought that null would be an ok data response here, along with the success status

src/store.rs
```diff
@@ -25,11 +28,40 @@ pub fn get_missing_files(root: &str, wanted: &[String]) -> io::Result<Vec<String
...
+    }
+
+    let path = file_path(root, hash)?;
+    let pathname = (path).to_str()
+        .map(String::from)
+        .unwrap();
+
+    if !file_exists(root, hash)? {
+        fs::create_dir_all(path.parent().unwrap())?;
+        fs::rename(temp_path, path)
```
richfitz, line 53: This is potentially problematic (either causing a failure or a copy and associated performance cost), depending on how temp_dir() is resolved because it _must_ be on the same file system for rename to work as expected. I believe that in outpack (or outpack.server) for this we used a temporary directory within the .outpack tree to enforce this.

This is a non-theoretical issue on at least some configurations of macOS where the temporary directory is at some user-specific mount which is a different logical file system and an atomic rename does not work.

Cargo.toml
```diff
@@ -23,6 +23,9 @@ lazy_static = "1.4.0"
 pest = "2.0"
 pest_derive = "2.0"
 itertools = "0.10.5"
+tempdir = "0.3.7"
+tempfile = "3.6.0"
+tar = "0.4.38"
```
richfitz, line 28: tar is just a development dependency though? can these be separated? actually - all three of these are just used in tests

richfitz (changes requested):

README.md
hillalex: sure, I guess you don't need it, it's just the standard convention for a POST request to return the location of the created resource (although come to think of it that would be the url `http://packit-instance/file/<hash>`)

src/store.rs
hillalex: So an issue for the case where this is used on metal as a cli? Ok, have changed to create a temp dir inside the outpack root...

README.md
richfitz: I think that returning a url would be fine, but I feel reasonably strongly we should not return a path that can't be read from

hillalex: Yeah, ok. Returning null and using the url to test the file has been uploaded.

src/store.rs
```diff
@@ -25,11 +29,37 @@ pub fn get_missing_files(root: &str, wanted: &[String]) -> io::Result<Vec<String
...
+}
+
+pub async fn put_file(root: &str, mut file: TempFile<'_>, hash: &str) -> io::Result<()> {
+    let temp_dir = tempdir_in(root)?;
+    let temp_path = temp_dir.path().join(hash);
+    file.persist_to(&temp_path).await?;
+    let alg = config::read_config(root)?.core.hash_algorithm;
+    let content = fs::read_to_string(&temp_path)?;
+    let valid_hash = hash::hash_data(content, alg);
+    if hash != valid_hash {
```
hillalex, line 42: Something just occurred to me; isn't it the case that the local user could legitimately be using a different hash algorithm to the remote? In which case we should include the algorithm in the request as well?

richfitz: yes, this is theoretically possible but there's heaps of support required to work out what would happen so won't happen practically for a while. The hash should be validated against the hash string (so parse the algorithm out of the hash and validate that, store the file in a path that matches that key, then *also* hard link a copy of our preferred algorithm's hash after reshashing). See https://github.com/mrc-ide/outpack/pull/77/files for some comments on that.

For now, consider checking if core.hash_algorithm matches the provided one and erroring if not compatible? This will work for all practical cases and be nicely future proof

src/hash.rs
```diff
@@ -49,6 +50,23 @@ pub fn hash_data(data: String, algorithm: HashAlgorithm) -> String {
     }
 }
 
+pub fn validate_hash(root: &str, hash: &str, content: &str) -> Result<(), Error> {
```
richfitz, line 53: thanks, this is great

richfitz (approved):

PR #20, opened 2023-06-26
Deserialize packet parameters and add helper function to test equality

r-ash:
This PR will
* Add a new packet into the example which has 4 parameters, 1 float, 1 int, 1 string and 1 bool
* Will deserailise these into an `Option<HashMap<String, serde_json::Value>>`
* Add a method `parameter_equals` which can be used for testing if a `Packet` has a parameter with a certain value (see #21 for an example of how I want to use this)

I want to use this to fix up #15 which we had to revert because the changes I made there were causing issues with deserialisation. See #18

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -20,14 +20,16 @@ pub struct Packet {
     pub parameters: Option<HashMap<String, serde_json::Value>>,
 }
 
-pub enum ParameterValue {
+#[allow(dead_code)]
+pub enum ParameterValue<'a> {
     Bool(bool),
-    String(String),
+    String(&'a str),
     Integer(i32),
     Float(f64)
 }
 
 impl Packet {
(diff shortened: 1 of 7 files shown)


src/metadata.rs
```diff
@@ -17,7 +17,52 @@ pub struct Packet {
     pub id: String,
     pub name: String,
     pub custom: Option<serde_json::Value>,
-    pub parameters: Option<serde_json::Value>,
+    pub parameters: Option<HashMap<String, serde_json::Value>>,
+}
+
+#[allow(dead_code)]
```
r-ash, line 23: These clippy warnings ignored atm but will be removed in #21

hillalex (approved):
nice, this seems like a good solution!

richfitz (approved):

PR #22, opened 2023-06-28
Remove the unpacked directory

richfitz:
This PR removes the `.outpack/unpacked` directory from the example, and updates the limited code that uses it to no longer need it.

See https://github.com/mrc-ide/outpack/pull/76 for context, this is where it was removed from outpack.

### DIFF
--- a/src/location.rs
+++ b/src/location.rs
@@ -70,7 +70,7 @@ pub fn read_locations(root_path: &str) -> io::Result<Vec<LocationEntry>> {
     Ok(packets)
 }
 
-pub fn get_local_location_id(root_path: &str) -> Result<String, Error> {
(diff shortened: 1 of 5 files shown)


hillalex (approved):
Looks good

src/location.rs
```diff
@@ -69,6 +70,16 @@ pub fn read_locations(root_path: &str) -> io::Result<Vec<LocationEntry>> {
     Ok(packets)
 }
 
+pub fn get_local_location_id(root_path: &str) -> Result<String, Error> {
```
hillalex, line 73:
```suggestion
pub fn get_local_location_id(root_path: &str) -> io::Result<String> {
```

(slightly less verbose way of writing this that we have used elsewhere)

richfitz: thanks!

PR #23, opened 2023-06-29
mrc-4310 POST metadata endpoint

hillalex:
Adds `POST /packet<hash>` which should replicate the logic of:

```
location_path_import_metadata <- function(str, hash, root) {
  meta <- outpack_metadata_load(as_json(str))
  id <- meta$id
  hash_validate_data(str, hash, sprintf("metadata for '%s'", id))

  unknown_files <- root_list_unknown_files(meta$files$hash, root)
  if (length(unknown_files) > 0) {
    stop(
      sprintf("Can't import metadata for '%s', as files missing:\n%s",
              id, paste(sprintf("  - %s", unknown_files), collapse = "\n")))
  }
  unknown_packets <- root_list_unknown_packets(meta$depends$packet, TRUE, root)
  if (length(unknown_packets) > 0) {
    stop(sprintf(
      "Can't import metadata for '%s', as dependencies missing:\n%s",
      id, paste(sprintf("  - %s", unknown_packets), collapse = "\n")))
  }

  if (!is.null(root$config$core$path_archive)) {
    dst <- file.path(root$path, root$config$core$path_archive,
                     meta$name, id, meta$files$path)
    root$files$get(meta$files$hash, dst)
  }

  writeLines(str, file.path(root$path, ".outpack", "metadata", id))
  local_id <- local_location_id(root)
  time <- Sys.time()
  mark_packet_known(id, local_id, hash, time, root)
}
```

(minus the path archive bit since remotes will never have a path archive)

### DIFF
--- a/schema/list.json
+++ b/schema/list.json
@@ -24,6 +24,6 @@
             }
         },
         "required": ["id", "name"],
-        "additionalProperties": false
+        "additionalProperties": true
     }
 }
--- a/schema/list.json
+++ b/schema/list.json
@@ -24,6 +24,6 @@
             }
         },
         "required": ["id", "name"],
-        "additionalProperties": true
+        "additionalProperties": false
     }
 }
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,4 +1,6 @@
 use std::ffi::{OsString};
+use std::time::UNIX_EPOCH;
+use cached::instant::SystemTime;
 use lazy_static::lazy_static;
 use regex::Regex;
 
@@ -15,8 +17,13 @@ pub fn is_packet_str(name: &str) -> bool {
     ID_REG.is_match(name)
 }
 
+pub fn time_as_num(time: SystemTime) -> f64 {
+    (time.duration_since(UNIX_EPOCH).unwrap().as_millis() as f64) / 1000.0
+}
+
 #[cfg(test)]
 mod tests {
+    use std::time::Duration;
     use super::*;
 
     #[test]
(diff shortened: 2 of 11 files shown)


schema/list.json
```diff
@@ -24,6 +24,6 @@
             }
         },
         "required": ["id", "name"],
-        "additionalProperties": false
```
hillalex, line 27: Because the `Packet` model has been extended, some extra properties are now being returned by the index endpoint (`files` and `depends`). At the moment the consumer of this endpoint (Packit) has no use for this info; this endpoint is just used to populate a more minimal index table. So the question is, should we introduce a separate struct (`PacketMinimal` or `PackitIndexItem` or something) to keep the index data returned more minimal?

hillalex: Another consideration is that at the moment, the logic to retrieve packet metadata is shared between the `/packet/metadata` and the querying and the latter will need the extra properties. So changing the model will also require duplicating that logic to some extent, just returning a more minimal struct in one case.

richfitz: I think we'll want separate schemas - one that is the exact properties as other outpack implementations (particularly as we build out more of what the rust version can do) - and others that are tailored to packit (or other internal uses).

tests/test_api.rs
```diff
@@ -415,6 +415,50 @@ fn file_post_handles_errors() {
     validate_error(&body, Some("invalid hash 'badhash'"));
 }
 
+#[test]
+fn can_post_metadata() {
+    let root = get_test_dir();
+    let rocket = outpack::api::api(root.clone());
+    let client = Client::tracked(rocket).expect("valid rocket instance");
+    let content = "{
```
richfitz, line 423: https://rahul-thakoor.github.io/rust-raw-string-literals/ this might be nicer for the embedded quotes

src/utils.rs
```diff
@@ -15,8 +17,13 @@ pub fn is_packet_str(name: &str) -> bool {
     ID_REG.is_match(name)
 }
 
+pub fn time_as_num(time: SystemTime) -> f64 {
+    (time.duration_since(UNIX_EPOCH).unwrap().as_millis() as f64) / 1000.0
```
richfitz, line 21: times are always a joy

richfitz (changes requested):
thanks - this looks good, but we should sort out the schema bits (and there's some conflicts sorry)

PR #26, opened 2023-08-02
Drop location id in favour of name

richfitz:
See https://github.com/mrc-ide/orderly2/pull/67 which pairs with this PR. This one should be merged first

> I was starting to move the schemas over, but realised that we never use this bit of indirection - and I don't even remember why it is here.
> 
> I think that the idea was that we could track back from a packet to the original location where it was generated (like a fingerprint) but that never made it through to the rest of the package. So now we just let the name be the primary key, which is what git does anyway.
> 
> From the tests you can see that the fallout is very minor - we really weren't using this "feature" at all for anything
>


src/config.rs
```diff
@@ -8,7 +8,6 @@ use strum_macros::{Display, EnumString};
 #[derive(Serialize, Deserialize, Debug)]
 pub struct Location {
     pub name: String,
```
hillalex, line 10: could possibly do away with this struct altogether if it's just a name?

richfitz: I think eventually we'll end up using some of the other fields if we ever want to implement pull etc (so we'd need type and args). A struct with just a string should be optimised down to just a string, so it doesn't cost much

PR #30, opened 2023-08-15
New implementation of hashing

richfitz:
This more closely matches the implementation in the python and R version and supports:

* hashing of both data and files
* a somewhat (but incompletely) rationalised error handling approach
* relaxes the use of confifg, which should not be checked here (but instead at server startup - to come, in a later PR)

The remains of an old fight with error types remains, but I think we can sort that out later. 

Cool things I found:

* The `From` trait for doing automatic conversions between error types (though you can't convert _from_ an error type you don't implement, but can convert to one)
* the `FromStr` trait for writing custom parse() handlers
* The `Display` trait, surprisingly easy to use
* serde makes it easy to change case on deserialise (with these three points, I removed the magic emum crate)


src/hash.rs
```diff
@@ -1,123 +1,299 @@
...
+        [1] "md5:81dc9bdb52d04dc20036dbd8313ed055"
+        > orderly2:::hash_data("1234", "sha1")
+        [1] "sha1:7110eda4d09e062aa5e4a390b0a572ac0d2c0220"
+        > orderly2:::hash_data("1234", "sha256")
+        [1] "sha256:03ac674216f3e15c761ee1a5e255f067953623c8b388b4459e13f978d7c846f4"
+        > orderly2:::hash_data("1234", "sha384")
+        [1] "sha384:504f008c8fcf8b2ed5dfcde752fc5464ab8ba064215d9c5b5fc486af3d9ab8c81b14785180d2ad7cee1ab792ad44798c"
+        > orderly2:::hash_data("1234", "sha512")
+        [1] "sha512:d404559f602eab6fd602ac7680dacbfaadd13630335e951f097af3900e9de176b6db28512f2e000b9d04fba5133e8b1c6e8df59db3a8ab9d60be4b97cc9e81db"
+         */
```
r-ash, line 224: Bit of old comment?

r-ash (approved):
This looks great

src/hash.rs
richfitz: this was showing where the values come from, just to explain the random string literals

src/hash.rs
```diff
@@ -1,123 +1,299 @@
...
-pub fn validate_hash(root: &str, hash: &str, content: &[u8]) -> Result<(), Error> {
-    let parsed = hash_parse(hash)?;
-    let alg = config::read_config(root)?.core.hash_algorithm;
-    if parsed.algorithm != alg.to_string() {
-        return Err(Error::new(ErrorKind::InvalidInput,
-                              format!("Hash algorithm {} does not match config. Expected {}",
-                                      parsed.algorithm, alg)));
+impl From<std::io::Error> for HashError {
+    fn from(e: std::io::Error) -> Self {
+        HashError::new(HashErrorKind::FileReadFailed, e.to_string())
```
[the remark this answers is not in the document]
richfitz: oh this is really nice

richfitz:
wes says go ahead and merge before he gets a chance to look at it, so let's do that to start munching through the backlog

PR #34, opened 2023-08-17
mrc-4336: Add support for querying with boolean and numeric types

r-ash:


src/query/query_types.rs
```diff
@@ -5,8 +5,17 @@ pub enum LookupLhs<'a> {
     Parameter(&'a str)
 }
 
+#[allow(dead_code)]
```
richfitz, line 8: what dead code are we getting here?

src/query/query_eval.rs
```diff
@@ -33,34 +33,74 @@ fn eval_latest<'a>(
...
+                _ => false
+            },
+            LookupLhs::Parameter(param_name) => packet.parameter_equals(param_name, &value)
         })
         .collect())
 }
 
+impl Packet {
+    #[allow(dead_code)]
+    fn parameter_equals(&self, param_name: &str, value: &LookupRhs) -> bool {
```
richfitz, line 57: this really begs for a comment or two explaining the idea I'm afraid. It might also be easier to read if the code arm of the match branches went into functions with descriptive names, and/or some helpers on the serde bit as it's really hard to scan atm

src/query/query.pest
```diff
@@ -10,17 +10,24 @@ expression         =  { expr }
...
 lookupName = { "name" }
 lookupParam = { "parameter:" ~ lookupParamName }
-lookupParamName = { ASCII_ALPHANUMERIC* }
+lookupParamName = @{ (ASCII_ALPHANUMERIC | "_" )+ }
+lookupValue = { string | boolean | float | integer }
 
+float = @{
+             "-"?
+             ~ ("0" | ASCII_NONZERO_DIGIT ~ ASCII_DIGIT*)
+             ~ ("." ~ ASCII_DIGIT*)
```
richfitz, line 27: not that we should , but numbers that this misses will include:

```
+5
1e-4
1e4
```

richfitz: Actually I don't know enough about how to read this to know how to parse it - does this exclude things with multiple dots in it? 

This stackoverflow answer looks like another approach? https://stackoverflow.com/a/68485567

src/query/query_parse.rs
```diff
@@ -184,4 +183,71 @@ mod tests {
...
+
+    #[test]
+    fn query_can_parse_parameters() {
+        let res = parse_query("parameter:x == \"foo\"").unwrap();
+        assert!(matches!(res, QueryNode::Lookup(LookupLhs::Parameter("x"), LookupRhs::String("foo"))));
+        let res = parse_query("parameter:x==\"foo\"").unwrap();
+        assert!(matches!(res, QueryNode::Lookup(LookupLhs::Parameter("x"), LookupRhs::String("foo"))));
+        let res = parse_query("parameter:longer==\"foo\"").unwrap();
+        assert!(matches!(res, QueryNode::Lookup(LookupLhs::Parameter("longer"), LookupRhs::String("foo"))));
+        let res = parse_query("parameter:x123==\"foo\"").unwrap();
```
richfitz, line 195: does 

```
        let res = parse_query("parameter:x123=='foo'").unwrap();
```

here work too?

richfitz (changes requested):

tests/test_query.rs
```diff
@@ -93,3 +93,70 @@ fn can_get_packet_by_parameter() {
...
+        outpack::query::run_query(root_path, "parameter:pull_data == \"true\"").unwrap();
+    assert_eq!(packets, "Found no packets");
+    let packets =
+        outpack::query::run_query(root_path, "parameter:pull_data == 1").unwrap();
+    assert_eq!(packets, "Found no packets");
+    let packets =
+        outpack::query::run_query(root_path, "parameter:pull_data == 0").unwrap();
+    assert_eq!(packets, "Found no packets");
+    let packets =
+        outpack::query::run_query(root_path, "parameter:pull_data == T");
```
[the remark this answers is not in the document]
r-ash: Way neater, thanks

src/query/query_types.rs
r-ash: Yep, no longer dead code now this PR actually uses it. Removed that

src/query/query.pest
r-ash: Yeah won't allow 1e4, deliberately left that out comment in the description. It wouldn't be too hard to add was just leaving it out for now to keep the already fairly confusing pattern as simple as possible. You can try it out here https://pest.rs/?g=N4Ig5gTghgtjURALhARwK4FMIE8AEBhhAvAcHgMoDyAkngH54BmANgPZQAuDeAorXgC%2BAHQB2Y1h26kAAsDFFFRYSAC0KgPwKlSxgAoVABhV4APngCCFAMI0aAfQByVRwC1eAJSr2AIjQDiNAAqPFa2Dn6BQQBUAJTaOkT6KgB0Joxhdr4BwXEJSoIgADQgAJaiAA7onMgghikAjABMAMwggkA#editor it excludes multiple dots. 

But I can add ability to write as +4, that should be fairly straightforward.

src/query/query_eval.rs
r-ash: Tried tidying that up a little by breaking the inner part (after we retrieve the parameter value) into another function. Sadly we're a little limited and can't break out each of the sections as we can't write a function signature which takes an enum variant! 

So I can't have e.g.

```
fn json_number_eq(json_val: serde_json::value::Value::Number, test_value: LookupRhs::Number) {
    if json_val.is_f64() {
        let test_number = serde_json::Number::from_f64(*test_val);
        match test_number {
            Some(number) => *json_val == number,
            None => false,
        }
    } else {
        *json_val == serde_json::Number::from(*test_val as i32)
    }
}
```

src/query/query.pest
r-ash: Ok actually done this now, collapsed the numbers to all be floats instead of having the float int distinction and then this wasn't too painful to add

r-ash:
@M-Kusumgar and @EmmaLRussell if you're wanting to take a look at this, some things which might be handy are the pest book https://pest.rs/book/intro.html, and also their online tool for playing with the grammar. There is a version [here](https://pest.rs/?g=N4Ig5gTghgtjURALhAelQAgMoFMcYBcALfABxwGcCMcATASwIHsIMiCDSKl1yqA6CBVQBiOoxYYAZpKgA7GlAoBPDAHcoq5hgp42TNQB056DPTmkArgQoYEZBLtoYARlqL1bkWPAjHjAI6WOBCqGOER4QC84cDYAPIAkhgAfjQAHqSsaQCiSRgAvsY4mawxAPpxFOZgADY4AGoI9FAu9QBilnIAxhgAPhhyTE0QLW04nT39ZnJS9Ok5pZTVTApFcsZDI2MdXb2RMRhxUnsAcrCUqRiGIAAUAJQ3hcbVcnWNza27U4fHZxe2NI3W5PNIlLLLeirK43R4gZ5ycEQSHQyIHWIZLIIk49c4wS5o9GVDC1KAESjUdak8lUQl037XEDUilPdbGcxzBZLCgrBQxY70IQEACCEDAVw580m3QIUIUaVqTCYAGtLKQmrVgtjBVRReL6UcSUrVaQEZL0tLZaiMPyMN0iAhgABGAA0GAATAUEcZFSq1REKnFfSbEs4BsG1Xj8OHjWqAAoIWAIiOkUM2w03ei0Vk%2B2OkKPpuI3OQXHNyFMJ6AwQuM0iJ-HkiBIUFGv2kSuwAtUvMdmAFmIAATit2FWAAwolEuVhQAZOMACWFpwAqgBZHIAJUSY%2BmN3KT3uAGpk3mNVrbVRRm9pi4lfV5NM5JYYC4Qt65E%2BX2-B8BjJEQSAAC0TwDDch6wgA-H%2BERpABAAMIEYKOE5Tqc8SnAAWpu8TlAAIokADiiQACpXMhk54YRJEAFSPAoMEYAB-AtuRU74URxG0VB9HhLBAB6Nw4C2AHgfCoFAZBZHjhR7EkYe9zcest5MPefIZiABAQMEiE3MRG7LjkOkgFIUC1LoRntLOWCGfC6yXjU6YACRFiAhhua58JpOYiLZIybk3GWphjqs5JyDYGBMFIOiaQ55jdJqDDXrQTCWOMGBBEwNLsh%2B34YEOdoOhA1EIkFIU4GFtiRdFV7imojBECl1DJal9TpZYmWUMY9oIAaRzQeEACEAH%2BR5u6uf5sJkacACaGCmKuZL2pc8iqN10Aym%2BJTdDgpDUE88jOG56gePadjIoMmU6GqpAsOSzjmG1ITKA5ZIwMY6wAOrziRORYHGwpjjk6bEk8RluQQYOGH4Yl%2Bf4tnGCALogOYVgEMgIB1lWOCNkg6Q2jE7r8HBIAFEAA#editor) preloaded with the current version of the grammar from this PR which you can play with

src/query/query.pest
```diff
@@ -10,17 +10,24 @@ expression         =  { expr }
 funcNames          = _{ latest }
 latest             =  { "latest" }
 
-infixExpression = { firstArg ~ infixFunction ~ secondArg }
+infixExpression = { firstArg ~ infixFunction ~ lookupValue }
```
r-ash, line 13: Changing the name here because then we get slightly better errors

src/query/query_types.rs
```diff
@@ -1,12 +1,19 @@
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
```
r-ash, line 1: We want PartialEq here for one of the tests, was passing this into a helper for testing and trying to use a match but turns out you can't match on a variable.. https://stackoverflow.com/a/72537694

src/query/query_eval.rs
```diff
@@ -33,34 +34,87 @@ fn eval_latest<'a>(
...
+    /// # Arguments
+    /// * `param_name` - A string slice that holds the name of the parameter to test
+    /// * `value` - A LookupRhs which holds the value to check equality for, can be a boolean, a
+    ///             string or a number
+    ///
+    /// # Return
+    /// * bool - true if the current packet has a parameter called `param_name` and its value
+    ///          is equal to the LookupRhs.
+    fn parameter_equals(&self, param_name: &str, value: &LookupRhs) -> bool {
+        if let Some(json_value) = self.get_parameter(param_name) {
```
EmmaLRussell, line 71: I stared at this for a bit wondering where `json_value` came from before realising this is a destructuring of the result of `self.get_parameter`!

src/query/query.pest
EmmaLRussell: Would it also be worth renaming `firstArg` to `lookup` or `lookupLhs` for the same reason?

EmmaLRussell (approved):
Thanks @r-ash, enjoyed seeing some real Rust in the wild. Don't really have much substantive to offer though! pest seems cool. 

Are we going to add inequality querying next?

src/query/query.pest
```diff
@@ -10,17 +10,24 @@ expression         =  { expr }
...
 firstArg        = { lookup }
-secondArg       = { string }
-infixFunction   = { char* }
+infixFunction   = { char{1, 2} }
```
M-Kusumgar, line 15: I wonder if this will cause problems? maybe something for the future but if you write `id = 21` it will have the infix function be chars "=" and "2" since you allow alphanumeric chars in char and then the lookup value will be "1" and this would be valid, there is a check in `query_parse` for this so not a big deal but i wonder if better definition of infixFunction allows us to better provide suggestions to the user like if it only captures "=" and then "21" instead of ["=", "2"] and then "1" we can probably say did you mean "==", "21" otherwise we it would be messy to find out the number was "21". Again definitely for the future!

M-Kusumgar (commented):
I don't have a huge amount to add about the code, all seems to work well and reads better especially after you replaced the `serde_json::value::Value` in the pattern matching! Couple comments about the parsing though, I don't know if these are intended or not but just something to be aware of i guess

src/query/query.pest
r-ash: Yes, can actually remove `firstArg` as a thing altogether. Tidied this up, and in the corresponding parsing code

r-ash:
> Are we going to add inequality querying next?

Yep along with a few other things
* Other tests `!=`, `<`, `<=`, `<`, `<=`
* Groupings, `(`, `!`, `&&` and `||`
* `single()`
* Subqueries & dependency searching

src/query/query.pest
```diff
@@ -10,17 +10,24 @@ expression         =  { expr }
...
+lookupParamName = @{ (ASCII_ALPHANUMERIC | "_" )+ }
+lookupValue = { string | boolean | number }
 
+number = @{
+    ("-" | "+")?
+    ~ ("0" | ASCII_NONZERO_DIGIT ~ ASCII_DIGIT*)
+    ~ ("." ~ ASCII_DIGIT*)?
+    ~ (^"e" ~ ("+" | "-")? ~ ASCII_DIGIT+)?
+}
+boolean = { "true" | "TRUE" | "false" | "FALSE" }
```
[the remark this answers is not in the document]
r-ash: Yeah good shout, missed this initially as we're just supporting R type TRUE/FALSE but I think after https://github.com/mrc-ide/orderly2/pull/28/files it should be fairly straight forward to support that. Any objections @richfitz ?

src/query/query.pest
r-ash: Ah good spot, actually would be ok with `id = 21` as it is sensitive to whitespace but would be a problem with `id=23` parsing as infix `=2` value `3`. Had left this bit deliberately relaxed and was managing the error handling in `query_parse` as you noticed. In general there is a bit of trickiness with this, do we use a really restrictive grammar and use the inbuilt pest errors (which are ok but not amazing) or do we have a less restrictive grammar but then return some more user-friendly errors in the parsing code. I've not thought that through properly yet as it is quite hard, so there is a bit of a mix here with this infixFunction being an example of a very permissive bit of a grammar. 

I think you're right though! It's more permissive than it needs to be, I've tightened that up for now.

M-Kusumgar (approved):

richfitz (approved):

PR #35, opened 2023-08-18
mrc-4541: Support querying with single quoted strings

r-ash:
This builds on #34 and should be merged after that

With this included you can run a query like `name == 'thing'` which will work. I've not fully thought through all the edge cases with using `"` or `\` inside your query as its complicated by how rust strings are oarsed in the testing and in the CLI so I'm punting on that. What this will give us is can run a query with a single quoted string and as long as it doesn't contain a `\` or `'` I think is should behave well.


richfitz (commented):
Pretty happy with ignoring escaped strings for now!

richfitz (approved):

PR #36, opened 2023-08-21
mrc-4558: Create an outpack root from the cli

richfitz:
This stretches our cli framework to the limit, it's getting a bit junky now, but the clap refactor can happen later.

This PR does a basic initialisation of an outpack root from the rust cli. This will simplify the deployment as we won't need another container, we can just use the server one we have - we might also update the api to create the directory as it starts, perhaps as an option.

Possible improvements would include indicating which configuration options have changed rather than just bailing.

Pretty pleased to have been able to do this with minimal googling now, stuck on a train crawling out of london...

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ FROM debian:bookworm-slim
 RUN  apt-get -yq update && \
      apt-get -yqq install openssh-client git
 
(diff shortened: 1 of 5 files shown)


r-ash (approved):
This also looks great

PR #37, opened 2023-09-07
mrc-4549: Add support for other comparison operators

r-ash:
This was a little bit of a fiddly one and there is some churn here but I think the end result isn't too bad actually!

This PR will add support to the querying language to run use additional tests `!=`, `<`, `>`, `<=`, `>=`. With this PR you should be able to run queries like
* `id != "123"`
* `name != "my_report"`
* `parameter:x > 2`
* `parameter:x >= 0.002`
* `parameter:y < 5`
* `parameter:y <= 5`

If you use a bool or a string on any of the ordering queries `<`, `<=`, `>`, `>=` this will return return no matches. We don't do any type coersion. This agrees with current verison in the R impl.

### DIFF
--- a/src/query/query.pest
+++ b/src/query/query.pest
@@ -11,7 +11,7 @@ funcNames          = _{ latest }
 latest             =  { "latest" }
 
 infixExpression = { lookup ~ infixFunction ~ lookupValue }
-infixFunction   = { ("=" | "!" | "<" | ">"){2} }
+infixFunction   = @{ ("=" | "!" | "<" | ">"){1,2} }
 
 lookup   = { lookupId | lookupName | lookupParam }
 lookupId = { "id" }
--- a/src/query/query_eval.rs
+++ b/src/query/query_eval.rs
@@ -226,90 +226,48 @@ mod tests {
         assert_eq!(params.get("pull_data").unwrap(),
                    &(serde_json::Value::Bool(true)));
 
-        assert!(packet.test_parameter("tolerance",
-                                      &Test::Equal,
-                                      &Literal::Number(0.001)));
-        assert!(!packet.test_parameter("tolerance",
-                                       &Test::Equal,
-                                       &Literal::Number(0.002)));
-        assert!(!packet.test_parameter("tolerance",
-                                       &Test::Equal,
-                                       &Literal::String("0.001")));
+        macro_rules! test_param {
+            ( $( $name:literal, $test:expr, $lit:expr => $result:expr )* ) => {
+                $(
+                assert_eq!(packet.test_parameter($name, $test, $lit), $result);
+                )*
+            };
+        }
 
-        assert!(packet.test_parameter("disease",
-                                      &Test::Equal,
-                                      &Literal::String("YF")));
-        assert!(!packet.test_parameter("disease",
-                                       &Test::Equal,
-                                       &Literal::String("HepB")));
-        assert!(!packet.test_parameter("disease",
-                                       &Test::Equal,
-                                       &Literal::Number(0.5)));
+        test_param!(
+            "tolerance", &Test::Equal, &Literal::Number(0.001)   => true
+            "tolerance", &Test::Equal, &Literal::Number(0.002)   => false
+            "tolerance", &Test::Equal, &Literal::String("0.001") => false
 
-        assert!(packet.test_parameter("size",
-                                      &Test::Equal,
-                                      &Literal::Number(10f64)));
-        assert!(packet.test_parameter("size",
-                                      &Test::Equal,
-                                      &Literal::Number(10.0)));
-        assert!(!packet.test_parameter("size",
-                                       &Test::Equal,
-                                       &Literal::Number(9f64)));
-        assert!(!packet.test_parameter("size",
-                                       &Test::Equal,
-                                       &Literal::Bool(true)));
+            "disease", &Test::Equal, &Literal::String("YF")   => true
+            "disease", &Test::Equal, &Literal::String("HepB") => false
+            "disease", &Test::Equal, &Literal::Number(0.5)    => false
 
-        assert!(packet.test_parameter("pull_data",
-                                      &Test::Equal,
-                                      &Literal::Bool(true)));
-        assert!(!packet.test_parameter("pull_data",
-                                       &Test::Equal,
-                                       &Literal::Bool(false)));
-        assert!(!packet.test_parameter("pull_data",
-                                       &Test::Equal,
-                                       &Literal::String("true")));
+            "size", &Test::Equal, &Literal::Number(10f64) => true
+            "size", &Test::Equal, &Literal::Number(10.0)  => true
+            "size", &Test::Equal, &Literal::Number(9f64)  => false
+            "size", &Test::Equal, &Literal::Bool(true)    => false
 
-        assert!(packet.test_parameter("tolerance",
-                                      &Test::NotEqual,
-                                      &Literal::Number(0.002)));
-        assert!(packet.test_parameter("tolerance",
-                                      &Test::LessThan,
-                                      &Literal::Number(0.002)));
-        assert!(packet.test_parameter("tolerance",
-                                      &Test::LessThanOrEqual,
-                                      &Literal::Number(0.002)));
-        assert!(packet.test_parameter("tolerance",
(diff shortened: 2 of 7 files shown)


src/query/query.pest
```diff
@@ -11,7 +11,7 @@ funcNames          = _{ latest }
 latest             =  { "latest" }
 
 infixExpression = { lookup ~ infixFunction ~ lookupValue }
-infixFunction   = { ("=" | "!" | "<" | ">"){2} }
+infixFunction   = @{ ("=" | "!" | "<" | ">"){1,2} }
```
r-ash, line 14: Can now be length 1 with `<` or `>`

src/query/query_eval.rs
```diff
@@ -7,7 +7,7 @@ use serde_json::value::Value as JsonValue;
 pub fn eval_query<'a>(index: &'a Index, query: QueryNode) -> Result<Vec<&'a Packet>, QueryError> {
     match query {
         QueryNode::Latest(inner) => eval_latest(index, inner),
-        QueryNode::Lookup(field, value) => eval_lookup(index, field, value),
+        QueryNode::Test(test, field, value) => eval_test(index, test, value, field),
```
r-ash, line 10: I've updated some of the naming here to be a bit more in line with the R impl and also I think makes it a bit clearer! Best perhaps to look at the query_types to understand this but essentially now there are `QueryNode`'s which are a `Test` which is one of =, !=, <, <=, >, >= and either side of this test can be a `Literal` (i.e. the was submitted on the command line) or a `Lookup` (name, id, parameter:x)

src/query/query.pest
EmmaLRussell: Would this also accept 2 charcter operators that aren't supported, like ">!"? ..Oh, looks like you get an error if you provide an invalid operator.

src/query/query_eval.rs
EmmaLRussell: Not sure what you mean by either side - that the Lookup and Literal can be on either side of the operator, but there should always be one of each? Or that you could have two Lookups e.g. to get results where param1 > param2? The signature of `eval_test` and the pest definition suggests that you've always got a lookup LHS and a literal RHS?

src/query/query_eval.rs
```diff
@@ -31,74 +31,93 @@ fn eval_latest<'a>(
...
+    /// * `test` - The type of test to run, ==, !=, <, <=, > or >=
+    /// * `value` - A Literal which holds the value to check equality for, can be a boolean, a
     ///             string or a number
     ///
     /// # Return
     /// * bool - true if the current packet has a parameter called `param_name` and its value
-    ///          is equal to the LookupRhs.
-    fn parameter_equals(&self, param_name: &str, value: &LookupRhs) -> bool {
+    ///          passes the specified test with the input test `value`.
+    fn test_parameter(&self, param_name: &str, test: &Test, value: &Literal) -> bool {
```
EmmaLRussell, line 105: This doesn't allow you to do < or > on strings to get inequality matches alphabetical order? Do we specifically want to not support that? I suppose there's not a obvious use case for it!

EmmaLRussell (commented):
Very cool to see a real use for various matches!

I wonder if we might eventually want to give some sort of feedback when there are no results because the query is invalid and couldn't possibly return matches because it uses an invalid type.. that might save a user some frustration!

src/query/query_eval.rs
r-ash: Sorry bit confusing language above. By either side I just meant in e.g. `a < b` that a and b are literally either side of the test `<`. But in the sentence above what I really wanted to say was that `Lookup` now has a different meaning to before. Before the `QueryNode::Lookup` was really a test that a field matched some value. So it would lookup the a value from the field name, then test it for equality. Now reworded this in the code so that it is a QueryNode::Test where one of the args is a new `Lookup` type and the other is a `Literal`.

You're right that at the moment the Lookup has to be on the LHS and the Literal on the RHS of the test. There is a (ticket)[https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4531/Make-query-lookups-work-in-any-order] to support the reverse order

src/query/query_eval.rs
r-ash: Yep that's right, have actively stopped that in the R impl too. Decided to turn it off just to avoid potential confusion from ambiguity of ordering. Is `"A" < "a"`? `"123" < "abc"`? `".." < "a"`? And are they consistent over langauges? In particular in Rust

```
fn main() {
    println!("{}", "a" < "A");
}
```
Gives `false` but in R

```
"a" < "A"
```
prints `TRUE` :fearful: 

So have deliberately not supported this for now but we could add it in future, but I am not sure how often people actually do this kind of comparison

richfitz (changes requested):
Agreed, this looks nice, the churn is easy enough to look past

src/query/query_eval.rs
richfitz: This is actually locale dependent

```r
x <- c("Bernard", "bernard", "béatrice", "Béatrice")
withr::with_locale(c(LC_COLLATE = "fr-FR"), sort(x))
withr::with_locale(c(LC_COLLATE = "C"), sort(x))
```

so not something we should encourage as it will differ based on people's machines!

src/query/query.pest
M-Kusumgar: i think if we are too strict with the infix function at the pest stage then we lose the ability to give more meaningful error messages, lets say someone wrote `=` instead of `==` or something like that, pest would throw its own error but if we allow it through pest, we could say "did you mean ==". I think keeping the infixFunction rule intentionally lax is probably a good thing

src/query/query_eval.rs
```diff
@@ -31,29 +31,63 @@ fn eval_latest<'a>(
...
+                },
+                JsonValue::String(string) => {
+                    Some(Literal::String(string))
+                },
+                _ => None // Parameters must be number, bool or string
+            }
+        } else {
+            None
+        }
+    }
```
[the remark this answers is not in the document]
M-Kusumgar: the error/option type propagation in this function feels a bit wonky to me, just want to make sure i understand, if the function fails to get a parameter it will just return a `None` but this silently goes up the chain and just returns `false` up the chain. Should we not give some sort of feedback to the user that the `param_name` doesnt exist?

Equally if it fails to parse a number as a f64 i feel like a lot has gone wrong there for it to silently return no matches to the user

M-Kusumgar (commented):
Looks great! just one comment really on error propagation but otherwise this is clean. Would be awesome if you get that macro working!

src/query/query_eval.rs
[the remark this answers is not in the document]
r-ash: On first point, we want this to be silent. If we run a query like `parameter:a == 2` this `get_parameter` function will be run for all packets in the archive which could be hundreds. Some might have that parameter but many won't. What we could do down the line is if this matches none of the packets we could then say something like `param_name` doesn't exist for any packet, or try detect typo in the parameter name or something.

On your second point, I think this should always return a Some. It runs this code https://docs.serde.rs/src/serde_json/number.rs.html#242 where the conversion is from a u64 or i64 to an f64. According to https://doc.rust-lang.org/reference/expressions/operator-expr.html#numeric-cast I think this will always work. If the integer has many digits some of these may be rounded. Seems very unlikely we will bump into this though

src/query/query_parse.rs
```diff
@@ -170,60 +175,77 @@ mod tests {
...
-        assert_query_node_lookup_number_eq(res, LookupLhs::Parameter("x"), 0.023);
+        assert_query_node_lookup_number_eq(res, Lookup::Parameter("x"), 0.023);
         let res = parse_query("parameter:x == -2.3e-2").unwrap();
-        assert_query_node_lookup_number_eq(res, LookupLhs::Parameter("x"), -0.023);
+        assert_query_node_lookup_number_eq(res, Lookup::Parameter("x"), -0.023);
+    }
+
+    #[test]
+    fn query_can_parse_tests() {
+        let res = parse_query("id == \"123\"").unwrap();
```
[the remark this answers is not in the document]
r-ash: Oh that looks good, yeah let's do that in 1 churn PR after this?

PR #41, opened 2023-09-28
mrc-4531: Enable comparison operators in any order

r-ash:
This builds on #40 and should be merged after that

After this PR you will be able to use comparison operators in any order i.e.
* `parameter:iso3 == "MWI"` or `"MWI" == parameter:iso3`
* `parameter:nmax < 2` or `2 > parameter:nmax`

### DIFF
--- a/src/query/query.pest
+++ b/src/query/query.pest
@@ -15,10 +15,12 @@ expr = _{ prefix? ~ (brackets | singleVariableFunc | noVariableFunc | infixExpre
 
 brackets = { "(" ~ body ~ ")" }
 
-noVariableFunc     =  { funcNames ~ "()" }
-singleVariableFunc =  { funcNames ~ "(" ~ body ~ ")" }
-funcNames          = _{ latest }
-latest             =  { "latest" }
+noVariableFunc          =  { noVariableFuncNames ~ "()" }
+singleVariableFunc      =  { singleVariableFuncNames ~ "(" ~ body ~ ")" }
+noVariableFuncNames     =  _{ latest }
+singleVariableFuncNames =  _{ latest | single }
+latest                  =  { "latest" }
+single                  =  { "single" }
 
 infixExpression = { lookup ~ infixFunction ~ lookupValue }
 infixFunction   = @{ ("=" | "!" | "<" | ">"){1,2} }
--- a/src/query/query.pest
+++ b/src/query/query.pest
@@ -22,7 +22,7 @@ singleVariableFuncNames =  _{ latest | single }
 latest                  =  { "latest" }
 single                  =  { "single" }
 
-infixExpression = { lookup ~ infixFunction ~ lookupValue }
+infixExpression = { lookup ~ infixFunction ~ lookupValue | lookupValue ~ infixFunction ~ lookup }
 infixFunction   = @{ ("=" | "!" | "<" | ">"){1,2} }
 
 lookup          = { lookupId | lookupName | lookupParam }
--- a/src/query/query_eval.rs
+++ b/src/query/query_eval.rs
@@ -1,13 +1,14 @@
-use std::collections::HashSet;
 use crate::index::Index;
 use crate::metadata::Packet;
 use crate::query::query_types::*;
 use crate::query::QueryError;
 use serde_json::value::Value as JsonValue;
+use std::collections::HashSet;
 
 pub fn eval_query<'a>(index: &'a Index, query: QueryNode) -> Result<Vec<&'a Packet>, QueryError> {
     match query {
         QueryNode::Latest(inner) => eval_latest(index, inner),
+        QueryNode::Single(inner) => eval_single(index, *inner),
         QueryNode::Test(test, field, value) => eval_test(index, test, value, field),
         QueryNode::Negation(inner) => eval_negation(index, *inner),
         QueryNode::Brackets(inner) => eval_brackets(index, *inner),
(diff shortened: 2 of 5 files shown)


src/query/query.pest
```diff
@@ -15,12 +15,14 @@ expr = _{ prefix? ~ (brackets | singleVariableFunc | noVariableFunc | infixExpre
...
-
-infixExpression = { lookup ~ infixFunction ~ lookupValue }
+noVariableFunc          =  { noVariableFuncNames ~ "()" }
+singleVariableFunc      =  { singleVariableFuncNames ~ "(" ~ body ~ ")" }
+noVariableFuncNames     =  _{ latest }
+singleVariableFuncNames =  _{ latest | single }
+latest                  =  { "latest" }
+single                  =  { "single" }
+
+infixExpression = { lookup ~ infixFunction ~ lookupValue | lookupValue ~ infixFunction ~ lookup }
```
richfitz, line 25: should this be

```suggestion
infixExpression = { (lookup | lookupValue)  ~ infixFunction ~  (lookup | lookupValue)}
```

(I may have the grouping operator wrong here!). So even though it makes no sense a user could write

```
latest(1 == 2)
```

which returns nothing useful ever, or

```
latest(parameter:a == parameter:b)
```

which is occasionally useful

r-ash: Hmm I could make that do able but have explicitly made it impossible (see tests here https://github.com/mrc-ide/outpack_server/pull/41/files#diff-41c5626f836cdc731bc0ed84cf86571988982366fa33facc8354a03f4e9660d6R473-R484)

Suppose that will just add a couple more cases to check in the evaluation - will take a look at that

richfitz: One please that this does turn up nontrivially, but hard to do much with in the rust version is `parameter:a == this:a` or `parameter:a == envir:a` which we use in quite a few places fetching dependencies in the R version - if we use this as the basis of the python support we will need that

r-ash: Ok I've given that a go now, you can compare different parameters and hopefully should be done in a way that we can add `envir` and `this` types in the future when we need them

src/query/query_eval.rs
```diff
@@ -69,88 +71,70 @@ fn eval_brackets<'a>(
...
+        }
+        (Test::Equal, Some(l), Some(r)) => l == r,
+        (Test::NotEqual, Some(l), Some(r)) => l != r,
+        (_, _, _) => false
     }
 }
 
 impl Packet {
-    pub fn get_parameter(&self, param_name: &str) ->  Option<Literal> {
+    pub fn lookup_value(&self, lookup: &Lookup) -> Option<Literal> {
```
richfitz, line 112: This looks like a nice generalisation of the previous version

src/query/query_eval.rs
```diff
@@ -69,88 +71,70 @@ fn eval_brackets<'a>(
...
+        }
+    }
+
+    pub fn get_parameter(&self, param_name: &str) -> Option<Literal> {
         if let Some(params) = &self.parameters {
             match params.get(param_name)? {
                 JsonValue::Number(number) => {
                     Some(Literal::Number(number.as_f64()?))
-                },
+                }
```
richfitz, line 125: Do we not need commas on matches? Mind blown

richfitz: I see - match blocks where we use `{}` don't care about commas, very odd

tests/test_query.rs
```diff
@@ -187,4 +187,19 @@ fn query_can_assert_single_return() {
...
+    let root_path = "tests/example";
+    test_query(root_path, "parameter:pull_data == TRUE", "20180220-095832-16a4bbed");
+    test_query(root_path, "TRUE == parameter:pull_data", "20180220-095832-16a4bbed");
+    test_query(root_path, "parameter:tolerance < 0.002", "20180220-095832-16a4bbed");
+    test_query(root_path, "0.002 > parameter:tolerance", "20180220-095832-16a4bbed");
+    test_query(root_path, "parameter:pull_data == parameter:pull_data", "20180220-095832-16a4bbed");
+    test_query(root_path, "parameter:tolerance < parameter:size", "20180220-095832-16a4bbed");
+    test_query(root_path, "parameter:missing < parameter:size", "Found no packets");
+    test_query(root_path, "2 == 1", "Found no packets");
+    test_query(root_path, "2 != 1",
```
richfitz, line 203: lol

richfitz (approved):

src/query/query_eval.rs
```diff
@@ -69,88 +71,70 @@ fn eval_brackets<'a>(
...
-        (test, Lookup::Parameter(param_name)) => packet.test_parameter(param_name, test, value),
+fn lookup_filter(packet: &Packet, test: &Test, lhs: &TestValue, rhs: &TestValue) -> bool {
+    let lhs_literal = match lhs {
+        TestValue::Literal(value) => Some(value.clone()),
+        TestValue::Lookup(lookup) => packet.lookup_value(lookup)
+    };
+    let rhs_literal = match rhs {
+        TestValue::Literal(value) => Some(value.clone()),
+        TestValue::Lookup(lookup) => packet.lookup_value(lookup)
+    };
```
EmmaLRussell, line 92: Tiny bit of repeated logic, is it worth pulling out into its own fn?

EmmaLRussell (approved):

M-Kusumgar (approved):
It does seem like the testing is going towards enum hell, like lengthwise. I tried looking for a way to use macros to shorten them but couldn't find a simple way, there is a way to just stick a string on to the end of an enum, like `Test::$s` i think but that requires error handling. Might be worth looking into though just to shorten this up.

src/query/query_eval.rs
r-ash: Thanks, I've tidied that up

PR #43, opened 2023-11-21
Remove the need for a preparsing step.

plietar:
The query language allows for a few short-form syntaxes that are only valid at the toplevel, not within complex expressions. Previously these were handled through a "preparse" step that would match on the string with regexes and convert it into a valid expression.

This commit replaces the preparsing step with a few new rules in the PEG-based parser. This makes the parser more consistent and should allow for better errorsi by keeping the spans accurate.

This also solves an inconvenience of the parser API. The nodes generated by the parser borrow from its input string, and are therefore bound by that lifetime. Previously that "input string" was actually the preprocessed query, rather than the original query. The output nodes would therefore have to borrow from that temporary string which made it impossible to write a high-level `parse<'a>(&'a str) -> QueryNode<'a>`. Instead the user of the API had to call the preparse and parse steps explicitly.

### DIFF
--- a/src/query/mod.rs
+++ b/src/query/mod.rs
@@ -10,7 +10,7 @@ extern crate pest;
 use crate::index::get_packet_index;
 use crate::query::query_eval::eval_query;
 use crate::query::query_format::format_query_result;
-use crate::query::query_parse::{parse_query, preparse_query, Rule};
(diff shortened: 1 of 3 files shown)


plietar:
Re-opened as #44. Turns out the CI throws a fit if the branch name has a slash in it.

PR #44, opened 2023-11-21
Remove the need for a preparsing step.

plietar:
The query language allows for a few short-form syntaxes that are only valid at the toplevel, not within complex expressions. Previously these were handled through a "preparse" step that would match on the string with regexes and convert it into a valid expression.

This commit replaces the preparsing step with a few new rules in the PEG-based parser. This makes the parser more consistent and should allow for better errorsi by keeping the spans accurate.

This also solves an inconvenience of the parser API. The nodes generated by the parser borrow from its input string, and are therefore bound by that lifetime. Previously that "input string" was actually the preprocessed query, rather than the original query. The output nodes would therefore have to borrow from that temporary string which made it impossible to write a high-level `parse<'a>(&'a str) -> QueryNode<'a>`. Instead the user of the API had to call the preparse and parse steps explicitly.

### DIFF
--- a/src/query/mod.rs
+++ b/src/query/mod.rs
@@ -10,7 +10,7 @@ extern crate pest;
 use crate::index::get_packet_index;
 use crate::query::query_eval::eval_query;
 use crate::query::query_format::format_query_result;
-use crate::query::query_parse::{parse_query, preparse_query, Rule};
+use crate::query::query_parse::{parse_query, Rule};
 use std::fmt;
 
 pub fn run_query(root: &str, query: &str) -> Result<String, QueryError> {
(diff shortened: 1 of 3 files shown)


r-ash (approved):
This looks great to me, think that simplifies parsing and nice to have it all in 1 function which we can call from elsewhere. I did have it like this originally and then split into a preparsing step but I think this is much nicer impl than what I had previously. I hadn't thought about the pain of borrowing from temporary string!

Have tagged @richfitz on this too if he wants to take a look also?

richfitz (approved):
LGTM

PR #45, opened 2023-11-21
Add parsing for `this:` and `environment:` in queries.

plietar:
These are not currently supported by the evaluation, since they only make sense when used from within a report script.

The outpack_query has a new `--parse-only` flag that can be used to see the result of the parser.

### DIFF
--- a/src/bin/outpack_query.rs
+++ b/src/bin/outpack_query.rs
@@ -1,18 +1,27 @@
 extern crate core;
 
 use getopts::Options;
-use std::{env, process};
+use std::{env, process::ExitCode};
+
+use outpack::query::QueryError;
 
 fn print_usage(program: &str, opts: Options) {
     let brief = format!("Usage: {} [options]", program);
     print!("{}", opts.usage(&brief));
 }
 
-fn parse_args(args: &[String]) -> (String, String) {
+enum Args {
+    Parse { query: String },
+    Eval { query: String, root: String },
+}
+
+fn parse_args(args: &[String]) -> Args {
     let program = args[0].clone();
     let mut opts = Options::new();
     opts.reqopt("q", "query", "outpack query (required)", "latest");
-    opts.reqopt("r", "root", "outpack root path (required)", ".");
+    opts.optopt("r", "root", "outpack root path", ".");
+    opts.optflag("", "parse-only", "parse the query without running it");
+
     let matches = match opts.parse(&args[1..]) {
         Ok(m) => m,
         Err(f) => {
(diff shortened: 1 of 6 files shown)


src/query/query_types.rs
```diff
@@ -1,12 +1,19 @@
...
     Parameter(&'a str),
 }
 
+#[derive(Debug, PartialEq)]
+pub enum Lookup<'a> {
+    Packet(PacketLookup<'a>),
+    This(&'a str),
+    Environment(&'a str),
+}
+
```
r-ash, line 16: Any thoughts on adding `this` and `environment` to the lookup instead of as a separate enum here. i.e. have

```
pub enum Lookup<'a> {
  Name,
  Id,
  Parameter(&'a str),
  This(&'a str),
  Environment(&'a str),
}
```

instead of as nested? 

I can see them logically being different things, were looking up from 3 different places. But wondering about that vs reducing some of the nesting. Just curious about why this way

src/bin/outpack_query.rs
```diff
@@ -1,39 +1,76 @@
...
     print!("{}", opts.usage(&brief));
 }
 
-fn parse_args(args: &[String]) -> (String, String) {
+enum Args {
+    Parse { query: String },
+    Eval { query: String, root: String },
+}
+
+fn parse_args(args: &[String]) -> Args {
```
r-ash, line 18: We've been chatting about using [clap](https://docs.rs/clap/latest/clap/) for our CLI interface, ticket here https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4553/Use-clap-for-the-outpack-rust-CLI . Something might be worth having a look at some point down the line if it looks interesting?

r-ash (approved):
This looks really great, one question just for my understanding

src/query/query_types.rs
plietar: My first attempt looked like your suggestion, then I had to update the [`Packet::lookup_value`](https://github.com/mrc-ide/outpack_server/blob/6afeb75482474724f8180235fa763be4f703709c/src/query/query_eval.rs#L106) function which is scoped for a single packet, and therefore This and Environment make no sense as arguments to that function. Having the nested hierarchy makes that evident.

src/bin/outpack_query.rs
plietar: Yes I saw that ticket, getopts is pretty limited so I think that would be a great idea. I kept the change to this file as low-effort as possible knowing it would probably get re-written soon.

src/query/query_types.rs
r-ash: ah yes that makes sense!

richfitz (approved):

richfitz:
Amazing how much `"{:?}"` can pull off - trying to imagine how much more work that would have been in C++

PR #46, opened 2023-11-24
Use clap for argument parsing.

plietar:
The various binaries are consolidated into a single `outpack` command, and their command line arguments are unified through the use of the clap crate.

Some of the error types around the crate are cleaned up a bit to allow this consolidation. In particular the api and init modules now use an `anyhow` error, and the query module uses a `thiserror` to keep a bit more precision. There are still a lot of (ab)uses of `io::Error` around, but these can be updated later.


src/bin/outpack/args.rs
```diff
@@ -0,0 +1,35 @@
...
+    /// Search for a packet in a repository
+    Search {
+        #[arg(short, long)]
+        root: String,
+        query: String,
+    },
+    /// Parse an outpack query, without evaluating it
+    Parse { query: String },
+    /// Start the outpack API server
+    ApiServer {
```
r-ash, line 31: How about just `Start`?

src/bin/outpack/args.rs
```diff
@@ -0,0 +1,35 @@
...
+        path: String,
+
+        #[arg(long)]
+        path_archive: Option<String>,
+
+        #[arg(long)]
+        use_file_store: bool,
+
+        #[arg(long)]
+        require_complete_tree: bool,
```
r-ash, line 20: Can these have docs on them too?

r-ash (changes requested):
This looks so nice, really tidy. clap looks really simple and error tidyup looks great.

Can you update the README? Has some usage docs in it with the current non-clap CLI.

src/bin/outpack/args.rs
plietar: The command is called `outpack`, so `outpack start` seems pretty ambiguous IMO.
Alternatively we could keep the old `outpack_server` command name, but then it makes the other commands a bit odd.

plietar: Gone with `StartServer`.

Cargo.toml
```diff
@@ -23,6 +23,9 @@ pest = "2.7"
 pest_derive = "2.7"
 itertools = "0.10.5"
 tempfile = "3.6.0"
+clap = { version = "4.4.8", features = ["derive"] }
```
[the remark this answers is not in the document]
plietar: Good catch. Thanks.

PR #47, opened 2023-11-28
Add python bindings for the query parser.

plietar:
The existing query parser is exposed as a new Python package called `outpack_query_parser`. The package contains a single function, `parse_query` which turns a string into an AST.

The AST is represented with Python objects, and can be printed and pattern matched on.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -12,18 +12,27 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - {os: ubuntu-latest}
-          - {os: macos-latest}
-          - {os: windows-latest}
+          - {os: ubuntu-latest,  py: '3.8'}
+          - {os: ubuntu-latest,  py: '3.9'}
+          - {os: ubuntu-latest,  py: '3.10'}
+          - {os: ubuntu-latest,  py: '3.11'}
+          - {os: macos-latest,   py: '3.11'}
+          - {os: windows-latest, py: '3.11'}
 
-    name: ${{matrix.config.os }}
+    name: ${{matrix.config.os }} (${{ matrix.config.py }})
     runs-on: ${{ matrix.config.os }}
+
     steps:
       - uses: actions/checkout@v2
-      - uses: actions-rs/toolchain@v1
+      - name: Set up Rust
+        uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
           components: clippy
+      - name: Set up Python ${{ matrix.config.py }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.config.py }}
       - name: Lint
         run: cargo clippy -- -D warnings
       - name: Format check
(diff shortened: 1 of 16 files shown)


plietar:
Some design decisions and open questions:
# WASM-based bindings
One alternative would be to export the parser as a wasm binary, which could be imported from Python.

Pros:
  - End-users don't need to have a Rust toolchain installed, assuming we build and provide the wasm files in CI.

Cons:
  - The tooling and ecosystem to do this is non-existent, tools like [wasm-bindgen](https://github.com/rustwasm/wasm-bindgen) are focused on binding to JS only, not Python.
  - No rich ABI, the parser would have to return the AST as a JSON blob. Not as convenient or idiomatic to use on the Python side, unless we add a "decoding" layer.
  - Pretty much all wasm implementations available in Python are written in Rust, so require a toolchain to build from source anyway. Binary wheels are available though.

I think as long as we produce binary wheels, then the approach of native bindings is superior to the wasm approach. The only downside is a bit of CI complexity, but maturin (the build tool used for Rust extenstions) [has good docs on this](https://www.maturin.rs/distribution).

# AST type definitions
Currently the parsed AST is returned as Python dataclass types. There's a few ways these types could be defined. I now have it where the Rust code calls the [`dataclasses.make_dataclass`](https://docs.python.org/3/library/dataclasses.html#dataclasses.make_dataclass) function on startup to create them. This used to be defined as native Rust types with enough magic to make it look like dataclasses (See https://github.com/mrc-ide/outpack_server/blob/f4b590d090b63790de764a5e95af7a1637db079a/src/query/python.rs#L64-L165) but it required quite a bit of boilerplate.

We could also give up on this and just return a JSON-like representation, though this wouldn't be as nice to process.

# Where should the bindings live
Instead of having the bindings in this repo, we could move them to https://github.com/mrc-ide/outpack-py (potentially as a sub-directory). I think I'm still on the fence for this once.

r-ash (approved):
Sorry this has taken me a while to get to, lots of new things here for me. This looks really cool and PyO3 looks excellent. Also cool to see maturin how you can switch the backend out with hatch, I think I need to read up a bit more on python packaging to understand everything here properly.

But this looks great, I like the idea of returning dataclasses. 

In terms of where this should live I wonder if another crate which depends on the query crate? Or maybe in another repo if we don't want to add hatch here.

.github/workflows/build-test.yaml
```diff
@@ -8,14 +8,31 @@ name: test
 
 jobs:
   build_and_test:
-    name: Rust project
-    runs-on: ubuntu-latest
+    strategy:
```
richfitz, line 11: can you enable prs into mrc-4553 to build please, that will test the python build

plietar: It was only targeting mrc-4553 while I was waiting for that to build. I've rebased it onto master now.

plietar:
> Could you enable the CI build on PRs into the target branch please so we can see the tests pass?

I've retargeted the PR to master so CI is running (and failing) now. Seems we'd never even tested the existing Rust code on WIndows until now, and there's a path-related test failure. I'll look into it tomorrow.

plietar:
Fixed the Windows path issues separately in https://github.com/mrc-ide/outpack_server/pull/48.

PR #49, opened 2023-12-11
mrc-4786: Build Python wheels in CI.

plietar:
This will allow users to install and import the Python bindings without needing to install a Rust toolchain.

The wheels should work on most mainstream platforms and architecture. We can add more if it turns out there's a gap somewhere. Additionally it uses abi3 wheels, which will work on any Python version >= 3.8, see https://pyo3.rs/v0.20.0/building_and_distribution#py_limited_apiabi3

The wheels are built unconditionally on every commit to master and PR commit. Additionally, whenever a tag is created the source package and wheels are pushed to PyPI.

I used TestPyPi to check the publish process and it seems to work fine. Of course there's a small chance this will not work on the first release attempt and need a bit more iteration later on.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -11,42 +11,28 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        config:
-          - {os: ubuntu-latest,  py: '3.8'}
-          - {os: ubuntu-latest,  py: '3.9'}
-          - {os: ubuntu-latest,  py: '3.10'}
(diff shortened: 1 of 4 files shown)


plietar:
I have a "pending trusted publisher" set up on PyPi (https://docs.pypi.org/trusted-publishers/creating-a-project-through-oidc/) waiting for us to make an initial tag. Once we do, it will create the package on PyPI under my ownership, and I'll be able to add more owners.

richfitz (approved):

PR #50, opened 2023-12-12
MRC 4770 - Add time field and remove custom from packit endpoint

absternator:
This PR alters data returned from enpoint `/packit/metadata?<known_since>` used by packit... This adds time field and removes custom field.

### DIFF
--- a/schema/server/list.json
+++ b/schema/server/list.json
@@ -17,10 +17,16 @@
                 "description": "Task parameters, used when running and for querying",
                 "type": ["null", "object"]
             },
-
-            "custom": {
-                "description": "Optional custom metadata, grouped under application keys",
-                "type": ["null", "object"]
+            "time": {
+                "description": "Information about the running time",
+                "start": {
+                    "description": "Time that the report was started, in seconds since 1970-01-01",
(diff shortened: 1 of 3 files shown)


schema/server/list.json
```diff
@@ -17,10 +17,16 @@
                 "description": "Task parameters, used when running and for querying",
                 "type": ["null", "object"]
             },
-
-            "custom": {
```
r-ash, line 21: Was custom not used for anything? Why remove it?

r-ash: Do we use PackitPacket anywhere else do you know? Or is that just for the API response?

absternator: Yup as discussed yesterday custom was not being used at all by Packit. And its only being removed from that api response object (PackitPacket)

richfitz: We might this back in at some point but OK to pull it out for now; this is where the orderly specific metadata would go

richfitz (approved):

r-ash (approved):
Thanks for chatting through this, all looking good as PackitPacket object only used for API to makes sense to drop the unused fields

PR #56, opened 2024-02-01
Use Path/PathBuf for the repository root.

plietar:
Using strings requires verbose conversions to and from path types anytime we want to use path operations (ie. `join()` is the vast majority of cases).

Using the path types consistently everywhere means we never have to worry about conversions. We get a `PathBuf` straight from our command-line arguments and stick to it. The only exception is tests, which frequently use a string literal to point to the example repo. These string literals now need to be replaced with `Path::new(...)`.

This is a fairly large but mostly mechanical change. `String` is replaced by `PathBuf` and `&str` is replaced by `&Path`. Conversion code here and there gets removed.


r-ash (approved):
Nice, that looks quite a bit tidier

PR #59, opened 2024-03-01
Add endpoint to run git fetch on outpack root

r-ash:
For building the packit runner we want to let users have the ability to update git so that in the runner UI they can see their latest branches. This is to address the use case
1. User runs a report
2. It errors for some reason
3. They fix it and push to github
4. They go to the UI and run git fetch and run the report again
5. It runs with updated sources

This PR is a lot of testing and a little impl. I've put the git setup helpers into a separate crate as it seems like this is the only way we can use them in integration tests and unit tests without having it exported by the outpack crate.

Hoping we can use the test setup here for
* Endpoint to return list of branches available so users can pick relevant branch they want to run on
* Endpoint to return info about default branch so we can show it in the UI for context for runners which have git switching turned off

Had quite a fight with the github actions which build the python bindings for the query parser. Turns out quite hard to have openssl available. I kind of wonder if we should instead break out the querying into a separate crate and just include that in the py03 build. The query parser doesn't need the git dependencies which are here and this adds a fair bit of complexity.


src/api.rs
```diff
@@ -156,6 +157,12 @@ async fn add_packet(
         .map(OutpackSuccess::from)
 }
 
+async fn git_fetch(root: State<PathBuf>) -> Result<OutpackSuccess<()>, OutpackError> {
```
richfitz, line 160: so this PR is a lot of work, but this is really nice

richfitz (approved):

src/git.rs
```diff
@@ -0,0 +1,51 @@
+use std::path::Path;
+
+use git2::Repository;
+
+pub fn git_fetch(root: &Path) -> Result<(), git2::Error> {
+    let repo = Repository::open(root)?;
+    let mut remote = repo
+        .find_remote("origin")
+        .expect("Failed to find remote 'origin'");
```
plietar, line 9: I guess these shouldn't happen in practice, but it would be good to return those errors rather than crash.

git-utils/README.md
```diff
@@ -0,0 +1,14 @@
+# git-utils
```
plietar, line 1: Would it make more sense for this to be called `test-utils` if the long term plan is to have more than just git stuff?

git-utils/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+        } else {
+            std::fs::copy(entry.path(), destination.as_ref().join(entry.file_name()))?;
+        }
+    }
+    Ok(())
+}
+
+fn create_file(repo_path: &Path, file_name: &str) {
+    let mut file = File::create(repo_path.join(file_name)).unwrap();
+    file.write_all(b"File contents").unwrap();
```
plietar, line 76:
```suggestion
    std::fs::write(repo_path.join(file_name), b"File contents").unwrap()
```

git-utils/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+    pub remote: Repository,
+    pub local: Repository,
+}
+
+// Initialise a git repo with a remote in the state that
+// remote - 3 commits, initial, first commit, second commit
+// local - 2 commits, initial, first commit
+// So that if we fetch on local then it should know about the second file
+pub fn initialise_git_repo(path: Option<&PathBuf>) -> TestGit {
+    let tmp_dir = TempDir::new("repo").expect("Temp dir created").into_path();
```
plietar, line 20: `into_path` means the directory never gets cleaned up, even after the program exists (unlike in R I think).
I would store the `TempDir` object inside the `TestGit` struct, then when that struct goes out of scope the folder is removed.

git-utils/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+    create_file(&remote_path, "new_file2");
+    git_add_all(&remote);
+    git_commit(&remote, "Second commit");
+
+    let default_branch = git_branch(&remote);
+    git_checkout(&remote, "other", true);
+    create_file(&remote_path, "new_file3");
+    git_add_all(&remote);
+    git_commit(&remote, "Third commit");
+    git_checkout(&remote, &default_branch, false);
```
plietar, line 46: As an alternative to working with the worktree and the index and having to switch branches around, you could also use the lower level APIs that just modify the contents of `.git` directly, eg. using https://docs.rs/git2/latest/git2/build/struct.TreeUpdateBuilder.html

Either options are fine, the lower level APIs are just a bit less stateful

git-utils/README.md
r-ash: Yes it would, I've updated that.

src/git.rs
r-ash: Thanks, fixed that

git-utils/src/lib.rs
r-ash: Ok had a quick look at that but a little confused about low level stuff. Tempted to stick with this since it is working but we can move to lower level if it proves brittle at all in the future?

plietar: Yeah that’s fine.

PR #60, opened 2024-04-03
Add a custom Python README.

plietar:
By default Hatch automatically includes the README.md file at the root
of the project as the Python package's description, which shows up when
looking at the package on PyPI. Since the README mostly contains
information about the server this could lead to confusion, with users
trying to start a server when what they really wanted was to be
redirected to the main outpack-py package.

This adds a minimal README with the Python specific information, and a
link to the main package. The new file is referenced from
`pyproject.toml`, ensuring this is what gets used as the description
rather than the default one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,15 +294,8 @@ The metadata should be written directly to the request body.
 
 ## Python bindings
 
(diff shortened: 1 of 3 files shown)


r-ash (approved):
Looks good!

PR #63, opened 2024-07-30
mrc-4973 Add list branches endpoint

M-Kusumgar:
This should add a get endpoint `/git/branches` to get information about the available branches. Main changes include:
* added git fetch and git branches endpoints to the README
* added `git_list_branches` function
* changed `git/fetch` endpoint to a get instead of a post, made more sense to me
* had to add `git config --global --add safe.directory /outpack` to `start-with-wait` because git was complaining about not being the owner of the repository

To test this manually, you can go to `/outpack` in the docker container and make some git branches or you can just mount a git repo into `/outpack` and then `curl localhost:8000/git/branches`

Note the original ticket also suggested having a boolean to show if the branch had unmerged commits or not, i could not find a super easy way to do this in rust (old logic for this was written in orderly.server in R by shelling out) so i have left it as a separate ticket: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5579/Add-key-to-determine-which-branch-has-unmerged-commits

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,13 +292,13 @@ The metadata should be written directly to the request body.
 }
 ```
 
-### GET /git/fetch
+### POST /git/fetch
 
-Does a git fetch on the repository (relevant for when runners clone down git repositories)
+Does a git fetch on the repository (relevant for when runners clone down git repositories). Expects an empty json body.
 
 ### GET /git/branches
 
-Returns an array of branches with their `name`, `commit_hash` (where branch pointer is), `time` (of last commit) and `message` (of last commit)
+Returns an array of branches with their `name`, `commit_hash` (where branch pointer is), `time` (of last commit) and `message` (of last commit in a string array split with respect to newline characters)
 
 #### Response
 
@@ -310,13 +310,13 @@ Returns an array of branches with their `name`, `commit_hash` (where branch poin
       "name": "main",
       "commit_hash": "ede307e23b2137ba2c7c3270e52f354f224942af",
       "time": 1722436575,
-      "message": "First commit\n"
+      "message": ["First commit"]
     },
     {
       "name": "other",
       "commit_hash": "e9078cf779584168c3781379380a3b1352545cda",
       "time": 1722436640,
-      "message": "Second commit\n"
+      "message": ["Second commit"]
     }
   ],
   "errors": null
--- a/README.md
+++ b/README.md
@@ -292,6 +292,37 @@ The metadata should be written directly to the request body.
 }
 ```
 
+### GET /git/fetch
+
+Does a git fetch on the repository (relevant for when runners clone down git repositories)
+
+### GET /git/branches
+
+Returns an array of branches with their `name`, `commit_hash` (where branch pointer is), `time` (of last commit) and `message` (of last commit)
+
+#### Response
+
+```
+{
+  "status": "success",
+  "data": [
+    {
+      "name": "main",
+      "commit_hash": "ede307e23b2137ba2c7c3270e52f354f224942af",
+      "time": 1722436575,
+      "message": "First commit\n"
+    },
+    {
+      "name": "other",
+      "commit_hash": "e9078cf779584168c3781379380a3b1352545cda",
+      "time": 1722436640,
+      "message": "Second commit\n"
+    }
+  ],
+  "errors": null
+}
+```
+
 ## Python bindings
 
 This crate provides Python bindings for its query parser. See
--- a/schema/server/branch.json
+++ b/schema/server/branch.json
@@ -1,0 +1,27 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "version": "0.0.1",
+  "type": "object",
+  "properties": {
+    "name": {
+      "oneOf": [
+        { "type": "null" },
+        { "type": "string" }
+      ]
+    },
+    "commit_hash": {
+      "type": "string"
+    },
+    "time": {
+      "type": "number"
+    },
+    "message": {
+      "oneOf": [
+        { "type": "null" },
+        { "type": "string" }
+      ]
+    }
+  },
+  "required": ["name", "commit_hash", "time", "message"],
+  "additionalProperties": false
+}
--- a/schema/server/branch.json
+++ b/schema/server/branch.json
@@ -4,10 +4,7 @@
   "type": "object",
   "properties": {
     "name": {
-      "oneOf": [
-        { "type": "null" },
-        { "type": "string" }
-      ]
+      "type": ["null", "string"]
     },
     "commit_hash": {
       "type": "string"
@@ -18,7 +15,12 @@
     "message": {
       "oneOf": [
         { "type": "null" },
-        { "type": "string" }
+        {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        }
       ]
     }
   },
--- a/schema/server/branch.json
+++ b/schema/server/branch.json
@@ -4,7 +4,7 @@
   "type": "object",
   "properties": {
     "name": {
-      "type": ["null", "string"]
+      "type": "string"
     },
     "commit_hash": {
       "type": "string"
@@ -13,15 +13,10 @@
       "type": "number"
     },
     "message": {
-      "oneOf": [
-        { "type": "null" },
-        {
-          "type": "array",
-          "items": {
-            "type": "string"
-          }
-        }
-      ]
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
     }
   },
   "required": ["name", "commit_hash", "time", "message"],
--- a/schema/server/branches.json
+++ b/schema/server/branches.json
@@ -1,0 +1,8 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "version": "0.0.1",
+  "type": "array",
+  "items": {
+    "$ref": "branch.json"
+  }
+}
--- a/src/api.rs
+++ b/src/api.rs
@@ -164,7 +164,9 @@ async fn git_fetch(root: State<PathBuf>) -> Result<OutpackSuccess<()>, OutpackEr
         .map(OutpackSuccess::from)
 }
 
-async fn git_list_branches(root: State<PathBuf>) -> Result<OutpackSuccess<Vec<git::BranchInfo>>, OutpackError> {
+async fn git_list_branches(
+    root: State<PathBuf>,
+) -> Result<OutpackSuccess<Vec<git::BranchInfo>>, OutpackError> {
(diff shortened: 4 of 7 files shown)


README.md
```diff
@@ -292,6 +292,37 @@ The metadata should be written directly to the request body.
...
+      "name": "main",
+      "commit_hash": "ede307e23b2137ba2c7c3270e52f354f224942af",
+      "time": 1722436575,
+      "message": "First commit\n"
+    },
+    {
+      "name": "other",
+      "commit_hash": "e9078cf779584168c3781379380a3b1352545cda",
+      "time": 1722436640,
+      "message": "Second commit\n"
```
richfitz, line 319: Will the embedded newlines work well here? I've had mixed success with doing this in json (dim memories of escaping issues), and I wonder if an array of strings might be better?

schema/server/branch.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "version": "0.0.1",
+  "type": "object",
+  "properties": {
+    "name": {
+      "oneOf": [
+        { "type": "null" },
+        { "type": "string" }
+      ]
```
richfitz, line 10:
```suggestion
      "type": ["null", "string"]
```

does this more compactly (& elsewhere)

schema/server/branches.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "version": "0.0.1",
+  "type": "array",
+  "items": {
+    "$ref": "branch.json"
```
richfitz, line 6: if we're not going to reuse branch.json elsewhere you might as well inline it here?

schema/server/branch.json
```diff
@@ -0,0 +1,27 @@
...
+    },
+    "commit_hash": {
+      "type": "string"
+    },
+    "time": {
+      "type": "number"
+    },
+    "message": {
```
richfitz, line 18: If you follow the suggestion above in the README this would become an array of strings (which might be zero length in the case of an empty message)

start-with-wait
```diff
@@ -15,5 +15,10 @@ do
         attempt_num=$((attempt_num++))
     fi
 done
+
+# git2 errors as the git repository mounted is not owned by
+# us, we need to declare it a safe directory to ignore ownership
+git config --global --add safe.directory /outpack
```
richfitz, line 21: who is it owned by?

src/api.rs
```diff
@@ -244,7 +252,8 @@ pub fn api(root: &Path) -> anyhow::Result<Router> {
         .route("/packit/metadata", get(get_metadata_since))
         .route("/file/:hash", get(get_file).post(add_file))
         .route("/packet/:hash", post(add_packet))
-        .route("/git/fetch", post(git_fetch))
+        .route("/git/fetch", get(git_fetch))
```
richfitz, line 255: why are we changing from `POST` to `GET` here? I thought we used POST on purpose for some reason (perhaps to indicate it changes the server contents)

richfitz (changes requested):
Suggestions throughout, take or leave as you feel is appropriate.

I'm not sure about the change from POST to GET - I feel that was deliberate but I don't recall the reasons

schema/server/branch.json
M-Kusumgar: done yh much better

README.md
M-Kusumgar: rip ill heed your warning, have changed it to array of strings

schema/server/branches.json
M-Kusumgar: have changed schema around in the next PR, branch.json does get reused so ill keep this as is for now

schema/server/branch.json
M-Kusumgar: okay ill still keep the null value to indicate message was not in utf-8 distinct from an empty message

src/api.rs
M-Kusumgar: there was no body just felt like get would do the job but think about it this is more in the flavour of a PUT request, but for now will revert it back

richfitz (approved):

README.md
```diff
@@ -292,6 +292,37 @@ The metadata should be written directly to the request body.
...
+#### Response
+
+```
+{
+  "status": "success",
+  "data": [
+    {
+      "name": "main",
+      "commit_hash": "ede307e23b2137ba2c7c3270e52f354f224942af",
+      "time": 1722436575,
```
plietar, line 312: I find RFC3339 timestamps nicer in a JSON context, but that's a matter of personal preference I guess.

Do we have precedent for using one or the other in our files?

plietar: Ok I looked it up and orderly's packet metadata uses Unix timestamps already. Bit of a shame IMO but it makes sense to stay consistent then.

schema/server/branch.json
plietar: You could also do a lossy conversion if you hit bad utf8 characters: https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8_lossy

src/git.rs
```diff
@@ -11,8 +12,48 @@ pub fn git_fetch(root: &Path) -> Result<(), git2::Error> {
...
+        commit_hash: branch_commit.id().to_string(),
+        time: branch_commit.time().seconds(),
+        message,
+    })
+}
+
+pub fn git_list_branches(root: &Path) -> Result<Vec<BranchInfo>, git2::Error> {
+    let repo = Repository::open(root)?;
+    let git_branches: Result<Vec<BranchInfo>, git2::Error> = repo
+        .branches(Some(BranchType::Local))?
```
plietar, line 47: As discussed separately, I think we actually want to list branches of the origin remote, not the local ones.

The local ones (if they even exist) will not be updated by the `fetch` API. Only the remote ones are (and these are actually the local view of the remote branches ...)

src/git.rs
```diff
@@ -11,8 +12,48 @@ pub fn git_fetch(root: &Path) -> Result<(), git2::Error> {
     Ok(())
 }
 
+#[derive(Serialize, Deserialize)]
+pub struct BranchInfo {
+    name: Option<String>,
```
plietar, line 17: I'm honestly surprised Git even allows this non-utf8 branch names.

I would be curious to see how GitHub behaves and what their APIs return on a non-utf8 branch name and/or message.

It might be worth ignoring these and simplifying the types and schema a bit.

plietar (changes requested):

src/git.rs
M-Kusumgar: yep done

src/git.rs
M-Kusumgar: sounds good, ive done the lossy conversion

src/git.rs
```diff
@@ -11,8 +12,54 @@ pub fn git_fetch(root: &Path) -> Result<(), git2::Error> {
...
+}
+
+fn get_branch_info(
+    branch_struct: Result<(Branch, BranchType), git2::Error>,
+) -> Result<BranchInfo, git2::Error> {
+    let branch = branch_struct?.0;
+    let lossy_name = String::from_utf8_lossy(branch.name_bytes()?);
+    let name = lossy_name
+        .strip_prefix("origin/")
+        .unwrap_or(&lossy_name)
```
plietar, line 30: Should we just ignore anything that doesn’t start with `origin/` instead?

src/git.rs
```diff
@@ -11,8 +12,54 @@ pub fn git_fetch(root: &Path) -> Result<(), git2::Error> {
...
+    })
+}
+
+pub fn git_list_branches(root: &Path) -> Result<Vec<BranchInfo>, git2::Error> {
+    let repo = Repository::open(root)?;
+    let git_branches: Result<Vec<BranchInfo>, git2::Error> = repo
+        .branches(Some(BranchType::Remote))?
+        // first branch seems to be HEAD, we don't want to display that to the
+        // users so skip it
+        .skip(1)
```
plietar, line 53: It seems preferable to filter by checking the value rather than by position. Otherwise you’re dependent on what could be an implementation detail.

plietar (approved):
Just some minor points left. I still don’t really understand the point of splitting the commit message by lines, but I don’t feel strongly about it either.

src/git.rs
M-Kusumgar: this should in theory never happen because we clone it ourselves, if it does happen for whatever reason, it would be helpful to see it in the UI and then we'll know the issue vs having it disappear and then working through the stack to debug why a branch isnt showing up, but im not too fussed if you would like me to ignore it!

src/git.rs
M-Kusumgar: yep thats better, have done it here instead, it was just easier to insert https://github.com/mrc-ide/outpack_server/pull/64/commits/b4c201a9e387ff391794c187ec148854debd5e24

M-Kusumgar:
> Just some minor points left. I still don’t really understand the point of splitting the commit message by lines, but I don’t feel strongly about it either.

this was because rich has had some issues in the past with `\n` characters and json encoding so he advised using an array of strings instead

src/git.rs
plietar: Sorry to be a bit nitpicky but it seems more appropriate to fix this here rather than in the next PR. I've double checked and it's more than just an implementation detail.

The result is actually sorted alphabetically, which means HEAD is generally first only because branch names are usually lowercase. If you have a branch that starts with a uppercase character in the A-G range it will get filtered out in place of HEAD.

src/git.rs
```diff
@@ -42,4 +89,22 @@ mod tests {
         let post_fetch_branches = git_remote_branches(&test_git.local);
         assert_eq!(post_fetch_branches.count(), 3); // HEAD, main and other
     }
+
+    #[test]
+    fn can_list_git_branches() {
+        let test_git = initialise_git_repo(None);
+        git_fetch(&test_git.dir.path().join("local")).unwrap();
+        let branches = git_list_branches(&test_git.dir.path().join("local")).unwrap();
+        let now_in_seconds = SystemTime::now()
```
plietar, line 98: This is going to be brittle, if the second rolls over between the call to `initialise_git_repo` and this line. I'm not sure it really is essential to test that value, but if you want to you'd have to do something like take the time before and after and check that it falls in the range.

M-Kusumgar: i dont mind just omitting the test for time value, feels fiddly with not much gain

src/git.rs
M-Kusumgar: no worries, that makes sense, ive copied the filter implementation from the next PR into here where i filter by name

plietar (approved):

PR #64, opened 2024-08-01
Mrc 5089 Add default branch info to list branches

M-Kusumgar:
This adds a `default_branch` key to the branch response. Main changes include:
* updated schema files for `branch-response.json` structure
* added a `default_branch` config value in a new `git` section of the config
* get default_branch if specified otherwise try `main` or `master` (in that order)

To test change the `git.default_branch` in `tests/example/config.json` to `null`, `"master"` or `"other"` and build and run using `buildkite/build` (comment out the pushes) and `buildkite/test` (you can curl `localhost:8000/git/branches`)


richfitz (commented):
before going into the details, there seems a key decision it would be good to check on here

tests/example/.outpack/config.json
```diff
@@ -17,5 +17,8 @@
       "type": "file",
       "args": []
     }
-  ]
+  ],
+  "git": {
```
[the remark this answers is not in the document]
M-Kusumgar: agreed, im gonna go for git block can be missing and/or `default_branch` potentially missing in which case main or master and not allow `null` for `default_branch`

src/git.rs
```diff
@@ -37,13 +48,39 @@ fn get_branch_info(
...
+pub fn git_list_branches(root: &Path) -> Result<BranchResponse, git2::Error> {
     let repo = Repository::open(root)?;
-    let git_branches: Result<Vec<BranchInfo>, git2::Error> = repo
+
+    let default_branch_name = read_config(root)
+        .map_err(io_err_to_git_err)?
+        .git
+        .default_branch;
+
+    let default_branch_struct = match default_branch_name {
```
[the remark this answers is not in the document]
M-Kusumgar: if im understanding this correctly, do you mean set `default_branch` in the `read_config` function to a default value? i think because we are trying both `main` and `master` it seems weird to choose one

src/git.rs
```diff
@@ -41,13 +52,38 @@ fn get_branch_info(
...
+pub fn git_list_branches(root: &Path) -> Result<BranchResponse, git2::Error> {
     let repo = Repository::open(root)?;
-    let git_branches: Result<Vec<BranchInfo>, git2::Error> = repo
+
+    let git_cfg = read_config(root).map_err(io_err_to_git_err)?.git;
+
+    let default_branch_struct = match git_cfg {
+        Some(GitConfig {
+            default_branch: Some(b),
+        }) => repo.find_branch(&b, BranchType::Local),
```
M-Kusumgar, line 67: I forgot you could do things like this in rust, best language ever

richfitz (approved):
This looks good to me - do you want to leave it for Paul to have a look too?

plietar:
Similar to the comment in #63 about using the remote's branches instead of the local ones. The default branch on the `origin` remote is probably what matters most, rather than guessing based on `main` or `master`.

You can use https://docs.rs/git2/latest/git2/struct.Remote.html#method.default_branch. This doesn't even need to be cached in the config (in fact probably it is probably best if it is not cached, in case it changes).

plietar (approved):

README.md
```diff
@@ -304,22 +304,30 @@ Returns an array of branches with their `name`, `commit_hash` (where branch poin
...
-      "time": 1722436575,
-      "message": ["First commit"]
+    "status": "success",
+    "data": {
+        "default_branch": {
+            "name": "main",
+            "commit_hash": "ede307e23b2137ba2c7c3270e52f354f224942af",
+            "time": 1722436575,
+            "message": ["First commit"]
+        },
```
plietar, line 314:
```suggestion
        "default_branch": "main",
```

src/git.rs
```diff
@@ -13,29 +13,35 @@ pub fn git_fetch(root: &Path) -> Result<(), git2::Error> {
...
+    let lossy_name = String::from_utf8_lossy(reference.name_bytes());
+    lossy_name
+        .strip_prefix("refs/remotes/origin/")
+        .unwrap_or_else(|| &lossy_name)
+        .to_string()
+}
+
+fn get_branch_info(branch: Branch) -> Result<BranchInfo, git2::Error> {
+    let git_ref = branch.get().resolve()?;
+    let name = get_branch_name(&git_ref);
```
plietar, line 39: Is this still needed? Seems like that was used when you were calling `get_branch_info` on `origin/HEAD`, but now that you don't anymore you can just keep the old code and do `branch.name_bytes()`.

PR #65, opened 2024-08-28
Add a streaming hash implementation.

plietar:
Previously files were being hashed by loading them in memory entirely first, then hashing the resulting buffer. When uploading large files, this puts a lot of memory pressure on the server.

Thankfully the hash types provide `io::Write`, meaning implementing a streaming hash is as straightforward as using `io::copy` from the input into them.

A few adjecent changes, mostly about wrapping some code behind a `spawn_blocking` call; a lot of our implementation is synchronous, and calling it directly from the main thread could cause it to block. The biggest offenders were the aforementioned hashing, and git operations.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -141,10 +141,13 @@ async fn add_file(
     hash: extract::Path<String>,
     file: Upload,
 ) -> Result<OutpackSuccess<()>, OutpackError> {
-    store::put_file(&root, file, &hash)
-        .await
-        .map_err(OutpackError::from)
-        .map(OutpackSuccess::from)
+    tokio::task::spawn_blocking(move || {
+        store::put_file(&root, file, &hash)
+            .map_err(OutpackError::from)
+            .map(OutpackSuccess::from)
+    })
+    .await
+    .unwrap()
 }
 
 async fn add_packet(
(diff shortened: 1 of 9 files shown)


src/api.rs
```diff
@@ -141,10 +141,13 @@ async fn add_file(
...
-        .await
-        .map_err(OutpackError::from)
-        .map(OutpackSuccess::from)
+    tokio::task::spawn_blocking(move || {
+        store::put_file(&root, file, &hash)
+            .map_err(OutpackError::from)
+            .map(OutpackSuccess::from)
+    })
+    .await
+    .unwrap()
```
M-Kusumgar, line 150: so this just moves it to another thread?

M-Kusumgar (approved):
Looks good to me, just a quick question to confirm im understanding that change correctly

src/api.rs
plietar: Yeah. Because the `store` module's API isn't async we need to be careful with calling functions that do heavy IO on it. I didn't wrap the more metadata-y ones on the basis that they probably won't block for that long.

The alternative would be to write everything async style and use `tokio::fs` everywhere instead of `std::fs`.

PR #66, opened 2024-08-28
Add process metrics.

plietar:
These allow to basic metrics about the process, such as CPU and memory usage.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,7 +89,7 @@ version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5ca11d4be1bab0c8bc8734a9aa7bf4ee8316d462a08c6ac5052f888fef5b494b"
 dependencies = [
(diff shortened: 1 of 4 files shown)


M-Kusumgar (approved):
wait thats all you have to do? i have never done these things but where does this info go? how does it know where to log this or is that a different thing and this just records it and stores it in a file?

plietar:
The way things work under the Prometheus model is that each service exposes a `/metrics` endpoint that returns the current value of the metrics. Prometheus periodically scrapes all those endpoints and stores it in its database.

https://github.com/mrc-ide/outpack_server/pull/53 did the initial ground work that added the endpoint.

See http://packit.dide.ic.ac.uk:9000/targets/outpack_server/malariaverse-sitefiles for what this looks like in practice. The Packit VM has an nginx reverse proxy that maps that forwards that path to `/metrics` on the outpack_server, so we only need to expose one port.

PR #67, opened 2024-08-28
Add an outpack_server_build_info metric.

plietar:
This describes the server's version and build environment. For now it only contains the version number (as specified in the `Cargo.toml`), the git revision and the Rust version used in the build.

The versions are listed as labels, and the metric has a value of 1. This follows the convention that is widespread in the Prometheus ecosystem.

This is what the resulting metric might look like:
```
outpack_server_build_info{revision="dd6b97db1f52ece5ea4b635ec43655d005ad7627",rustc="1.78.0",version="0.3.0"} 1
```

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -226,6 +226,23 @@ pub fn register_process_metrics(_registry: &Registry) -> prometheus::Result<()> 
     Ok(())
 }
 
+pub fn register_build_info_metrics(registry: &Registry) -> prometheus::Result<()> {
+    let opts =
+        Opts::new("build_info", "Build information about the server").namespace("outpack_server");
+    let labels = ["version", "revision", "rustc"];
+    let values = [
+        env!("CARGO_PKG_VERSION"),
+        env!("VERGEN_GIT_SHA"),
+        env!("VERGEN_RUSTC_SEMVER"),
+    ];
+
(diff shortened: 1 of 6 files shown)


src/metrics.rs
```diff
@@ -226,6 +226,23 @@ pub fn register_process_metrics(_registry: &Registry) -> prometheus::Result<()>
...
 }
 
+pub fn register_build_info_metrics(registry: &Registry) -> prometheus::Result<()> {
+    let opts =
+        Opts::new("build_info", "Build information about the server").namespace("outpack_server");
+    let labels = ["version", "revision", "rustc"];
+    let values = [
+        env!("CARGO_PKG_VERSION"),
+        env!("VERGEN_GIT_SHA"),
+        env!("VERGEN_RUSTC_SEMVER"),
```
M-Kusumgar, line 236: these are set by default im assuming?

M-Kusumgar (approved):
looks good to me

src/metrics.rs
plietar: `CARGO_PKG_VERSION` is set automatically by Cargo. The `VERGEN_*` ones are set by the build script, using the vergen crate.

Note that `env!` looks up environment variables at build time, not runtime.

PR #68, opened 2024-11-22
Correct the documentation of the packit/metadata endpoint

david-mears-2:
Here is the start of a response I got:

```
curl -s http://localhost:8000/packit/metadata | jq

{
  "status": "success",
  "data": [
    {
      "id": "20240729-154633-10abe7d1",
      "name": "artefact-types",
      "parameters": null,
      "time": {
        "start": 1722267993.0676,
        "end": 1722267993.0971
      }
    },
    {
      "id": "20240729-154635-88c5c1eb",
      "name": "computed-resource",
      "parameters": null,
      "time": {
        "start": 1722267995.5361,
        "end": 1722267995.5652
      }
    },
 ```
 
 Not sure if this is documented elsewhere too

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,14 +121,20 @@ e.g. `/packit/metadata?known_since=1683117048`.
         {
             "id": "20220812-155808-c873e405",
             "name": "depends",
-            "custom": { "orderly": { "display": "Report with dependencies" }}
-            "parameters": null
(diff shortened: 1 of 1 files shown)


david-mears-2:
Instead of changing the docs we'll change the system to match the spec.

PR #69, opened 2024-11-26
Add custom metadata to /packit/metadata endpoint

david-mears-2:
This brings the implementation of the API to be more in line with the spec as it was in the README: namely, the /packit/metadata endpoint will now pass on the 'custom' property as part of the response. (The README also didn't show that the 'time' property was being returned.)

### DIFF
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -278,6 +278,13 @@ async fn can_list_metadata() {
         1503074938.2232
     );
     assert_eq!(
+        entries[0].get("custom").unwrap().as_object().unwrap()
+            .get("orderly").unwrap().as_object().unwrap()
+            .get("artefacts").unwrap().as_array().unwrap()[3]
+            .get("description").unwrap().as_str().unwrap(),
+        "Projected Coverage for routine immunisation in PINE countries"
+    );
+    assert_eq!(
         entries[1].get("id").unwrap().as_str().unwrap(),
         "20170818-164847-7574883b"
     );
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -278,23 +278,7 @@ async fn can_list_metadata() {
         1503074938.2232
     );
     assert_eq!(
-        entries[0]
-            .get("custom")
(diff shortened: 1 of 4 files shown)


plietar (approved):
LGTM. A minor suggested cleanup about the test, up to you if you want to change it or keep it as is to keep consistency with the rest of the file. I'm pretty ambivalent about it.

plietar:
Did GitHub just forget about the code review comment I had written or something?? Sorry I will write it again

tests/test_api.rs
```diff
@@ -277,6 +277,26 @@ async fn can_list_metadata() {
...
+            .get("artefacts")
+            .unwrap()
+            .as_array()
+            .unwrap()[3]
+            .get("description")
+            .unwrap()
+            .as_str()
+            .unwrap(),
+        "Projected Coverage for routine immunisation in PINE countries"
+    );
```
plietar, line 299: I know the rest of the file is written in this style and you are just following that, but that is way more verbose than it needs to be.

```rust
assert_eq!(entries[0]["custom"]["orderly"]["artefacts"][3]["description"], "Projected Coverage ...");
```

`serde_json::Value` implements the Index trait for both integers and string keys. If the value isn't of the right kind, the indexing returns a Null variant and the equality assertion would eventually fail. It also has a `PartialEq<&str>` implementation, which means that you shouldn't need the final `.as_str().unwrap()` call.

Alternatively, there is also the `.pointer` method, eg.

```rust
entries[0].pointer("/custom/orderly/artefacts/3/description")
```

Pointer syntax comes from https://www.rfc-editor.org/rfc/rfc6901

https://docs.rs/serde_json/latest/serde_json/value/enum.Value.html#impl-Index%3CI%3E-for-Value
https://docs.rs/serde_json/latest/serde_json/value/enum.Value.html#method.pointer

david-mears-2: Oh, I hadn't seen pointer syntax before, that's handy!

I will use the index syntax.

PR #71, opened 2025-01-28
Remove git support.

plietar:


M-Kusumgar (approved):
love it when code is killed

PR #73, opened 2025-02-11
Report an error when pushing conflicting metadata.

plietar:
Our packet IDs are generally assumed to be globally unique, but this is not guaranteed by construction: the ID is not a hash of its contents, and the metadata store is not content-addressed.

This means one could try to upload two different packets to the store with the same ID. Previously, outpack server would accept the second request but skip writing to the file. It is better to detect this situation and to return an HTTP error when it occurs.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -2,7 +2,9 @@ use cached::instant::SystemTime;
 use lazy_static::lazy_static;
 use regex::Regex;
 use std::ffi::OsString;
+use std::path::Path;
 use std::time::UNIX_EPOCH;
+use std::{fs, io, io::Write};
 
 lazy_static! {
     static ref ID_REG: Regex = Regex::new(r"^([0-9]{8}-[0-9]{6}-[[:xdigit:]]{8})$").unwrap();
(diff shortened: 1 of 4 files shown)


src/utils.rs
```diff
@@ -20,10 +22,34 @@ pub fn time_as_num(time: SystemTime) -> f64 {
     (time.duration_since(UNIX_EPOCH).unwrap().as_millis() as f64) / 1000.0
 }
 
+/// Write a byte slice to disk.
+///
+/// Succeeds if the file already exists with identical contents.
+/// On the other hand, an AlreadyExists error is returned if the file exists with different contents.
+pub fn write_file_idempotent(path: &Path, contents: &[u8]) -> io::Result<()> {
```
plietar, line 29: I don't love this name, but I also couldn't find a better one that expresses the idea of detecting conflicts.

src/utils.rs
```diff
@@ -20,10 +22,34 @@ pub fn time_as_num(time: SystemTime) -> f64 {
...
+        .create_new(true)
+        .open(path)
+    {
+        Ok(mut f) => {
+            f.write_all(contents)?;
+        }
+        Err(err) if err.kind() == io::ErrorKind::AlreadyExists => {
+            if fs::read(path)? != contents {
+                return Err(err);
+            }
```
richfitz, line 41: Are we ok with end-of-line standardisartion between windows/everyone else here? I can't remember what we do in this case - perhaps we dodge it by spitting everything out on one line first?

plietar: outpack_server doesn't ever do any special processing of newline or anything else. Metadata files are an immutable blob of bytes.

PR #76, opened 2025-06-09
Add additional metadata

richfitz:
We need a location entry for orderly2 to be able to interact with the outpack archive, at least for things like migrating data.  This PR adds the "local" location, which orderly treats specially (to the point where it cannot add it).

Metadata on initialisation, after this PR:

```
{
  "core": {
    "hash_algorithm": "sha256",
    "path_archive": null,
    "use_file_store": true,
    "require_complete_tree": false
  },
  "location": [
    {
      "name": "local",
      "type": "local",
      "args": {}
    }
  ]
}
```

Currently the `location` section is the empty list.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "outpack"
-version = "0.3.2"
+version = "0.3.3"
 edition = "2021"
 rust-version = "1.70"
 build = "build.rs"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "outpack"
-version = "0.3.3"
+version = "0.3.2"
 edition = "2021"
 rust-version = "1.70"
 build = "build.rs"
--- a/src/config.rs
+++ b/src/config.rs
@@ -7,12 +7,21 @@ use std::result::Result;
 use crate::hash::HashAlgorithm;
 
 #[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct Empty {}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
 pub struct Location {
     // Practically, doing anything with locations (therefore needing
     // access to the "type" and "args" fields) is going to require we
     // know how to deserialise into a union type; for example
     // https://stackoverflow.com/q/66964692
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "outpack"
-version = "0.3.2"
+version = "0.3.3"
```
plietar, line 3: This is unrelated to this PR, but in the past I've done explicit PRs for bumping the version number, following https://github.com/mrc-ide/outpack_server?tab=readme-ov-file#releasing

I would be fine with using the same workflow we do in our R stuff, and bump the version on every commit. We'd want to tweak our CI to a) ensure the version has been bumped and b) push to PyPI on every push to main, not just on tags.

src/config.rs
```diff
@@ -86,6 +100,9 @@ mod tests {
     #[test]
     fn can_write_config() {
         let cfg = Config::new(None, true, true).unwrap();
+        assert_eq!(cfg.location.len(), 1);
+        assert_eq!(cfg.location[0].name, String::from("local"));
+        assert_eq!(cfg.location[0].loc_type, String::from("local"));
```
plietar, line 105:
```suggestion
        assert_eq!(cfg.location[0].name, "local");
        assert_eq!(cfg.location[0].loc_type, "local");
```

does that not work?

src/config.rs
```diff
@@ -6,13 +6,22 @@ use std::result::Result;
...
     // access to the "type" and "args" fields) is going to require we
     // know how to deserialise into a union type; for example
     // https://stackoverflow.com/q/66964692
+    //
+    // However, we need to support the 'local' type, which takes no
+    // arguments, so implement enough here to be able to write one.
     pub name: String,
+    #[serde(rename = "type")]
+    pub loc_type: String,
+    pub args: Empty,
```
plietar, line 24:
```suggestion
    pub args: HashMap<String, serde_json::Value>,
```

maybe?

The Empty class feels a bit weird, especially if it is ever going to de-serialize a non-empty thing. I guess serde probably ignores unknown fields so it could be fine as is. Up to you

plietar (approved):

PR #77, opened 2025-06-24
Add docker gha workflow

richfitz:
Based on the build in acme-buddy.  I have removed and disabled the buildkite build, and there is no simultaneous push to dockerhub.


EmmaLRussell (commented):
So with this set up, you have the "push" action run pushing an image tagged with the branch name: 
```
#17 pushing ghcr.io/mrc-ide/outpack_server:mrc-6567 with docker
```
...and the "pull_request` action pushing an image tagged with the pr number:
```
#17 pushing ghcr.io/mrc-ide/outpack_server:pr-77 with docker
```

..which is a bit different from our usual system of images tagged with branch name, and also one confusingly tagged with the merge commit (about which there was considerable discussion yesterday... !)

I guess that's fine if we're really just interested in the branch!

